PR #9, opened 2021-05-17
ewc-media-controls (WIP)

benjaminARowe:
## What
Initial version of ewc-media-controls. This component can be applied to any element that extends HTMLMediaElement

## Why
This component adds support for play/pause of HTMLAudioElement and (eventually) ewc-spectrogram, and user configurable back / forward buttons

## How
Using getElementById to get the element associated with the 'for' property on ewc-media-controls, then checking if the element is an instance of HTMLMediaElement. Error's are thrown if the element does not exist or is not an instance of HTMLMediaElement. Warning is logged to console if no element id is provided

## Testing (WIP)
Added unit tests for:
    - Warns and renders if no media source is given
    - Throws an error when the media element does not exist
    - Throws an error when the element is not a HTMLMediaElement
    - Renders if the element is a HTMLMediaElement
    - Renders a back button if onBack is defined
    - Renders a back and forward buttons if onBack and onForward are defined
(WIP) some tests fail if instanceof HTMLMediaElement is used. Currently using a workarround. The following solutions were attempted:
    - Use stencil's built-in mock window (newSpecPage.win)
        - Stencil's implementation of window is incomplete, so this fails
    - Use "jsdom"
        - jsdom is incompatible with stencil in some ways relating to limited support for shadowdom. e.g. querySelector("audio")[0] instanceof HTMLMediaElement -> false
    - Add "allow-unsafe" id to audio for tests
        -   This works, but is a bad solution


src/components/media-controls/media-controls.tsx
```diff
@@ -0,0 +1,79 @@
...
+  @Prop() back: string = 'Back';
+  @Prop() play: string = 'Play';
+  @Prop() pause: string = 'Pause';
+  @Prop() forward: string = 'Forward';
+
+  // Action to be performed when the back button is pressed
+  @Prop() onBack: VoidFunction;
+
+  // Action to be performed when the forward button is pressed
+  @Prop() onForward: VoidFunction;
```
atruskie, line 23: These should be `@Event`s no? With the type `EventEmitter`?

src/components/media-controls/media-controls.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { getMedia } from '../../utils/media';
+
+@Component({
+  tag: 'ewc-media-controls',
+  styleUrl: 'media-controls.css',
+  shadow: true,
+})
+export class MediaControls {
+  // The id of the HTMLMediaElement
+  @Prop() for: string;
```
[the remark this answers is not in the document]
atruskie: Ok I'm reading the docs, `@prop`s that are not simple types are not output by the compiler. I'd still like to see the same property used for the complex and string types where possible/makes sense.

src/components/media-controls/media-controls.tsx
```diff
@@ -0,0 +1,79 @@
...
+  };
+  pauseSource = () => {
+    if (this.mediaSource) {
+      this.mediaSource.pause();
+    }
+  };
+  toggleSource = () => {
+    if (this.mediaSource) {
+      this.mediaSource.paused ? this.playSource() : this.pauseSource();
+    }
```
atruskie, line 67: Overly defensive here?

If not, collapse to null-access operator chain?

src/components/media-controls/readme.md
```diff
@@ -0,0 +1,19 @@
...
+<!-- Auto Generated Below -->
+
+
+## Properties
+
+| Property | Attribute | Description     | Type     | Default     |
+| -------- | --------- | --------------- | -------- | ----------- |
+| `first`  | `first`   | The first name  | `string` | `undefined` |
+| `last`   | `last`    | The last name   | `string` | `undefined` |
+| `middle` | `middle`  | The middle name | `string` | `undefined` |
```
atruskie, line 14: Update

src/utils/media.ts
```diff
@@ -0,0 +1,29 @@
+export function MediaException(message) {
+  this.message = message;
+  this.name = 'MediaException';
+}
```
atruskie, line 4: This should be a type that extends `Error`.

Also should be namespaced

src/utils/media.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Get the media element when the component loads or is updated
+ */
+export function getMedia(element: string) {
+  let mediaSource = document.getElementById(element);
```
atruskie, line 10: global reference here is *not* ok.

Pass in the `<host/>` element and use it's query selector API

src/utils/media.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  // Warn if no element has been set
+  if (!element) {
+    console.warn('No HTMLMediaElement was defined');
```
atruskie, line 14: No global access is allowed to `console`. Inject it.

Also every console message generated by this project should be namespaced. Make a global helper service.,

src/utils/media.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+  // If the HTMLELement does not throw an error
+  else if (!mediaSource) {
+    throw new MediaException(`The HTMLElement ${element} does not exist`);
+  }
+  // If mediaSource HTMLElement is a HTMLMediaElement
+  else if (mediaSource instanceof window.HTMLMediaElement || mediaSource.id == 'allow-unsafe') {
```
atruskie, line 21: again no global variable access allowed

atruskie (changes requested):
- There's a lot of noise in this PR. Try to keep branches clean with just  the relevant changes.
- For example, all the stub definitions would have been better placed in another PR

For the rest of this review, I'm ignoring all definitions other than the media controls. Treat the media controls as a template for which we'll base all the other stubs.

I'm also going to suggest making it standard practice to create an issue describing the work you are going to undertake before you start - to make sure we're on the same page before you start.

Does the current build pipeline support scss/sass compilation? There is a `stencil-sass` plugin?

I didn't see any tests for CSS customisation. Add a test that checks the icons can be replaced by a host style. e.g. `ewc-media-controls::part(pause) { color: red; }` 

The most important part of this PR will be establishing some kind of style guide afterwards.

I'm going to require that you turn on the no global variable access option in your linter.

src/components/media-controls/media-controls.tsx
```diff
@@ -0,0 +1,79 @@
...
+  addPlayPauseListener = () => {
+    if (this.mediaSource) {
+      this.mediaSource.addEventListener('pause', () => {
+        this.mediaPaused = this.mediaSource.paused;
+      });
+      this.mediaSource.addEventListener('play', () => {
+        this.mediaPaused = this.mediaSource.paused;
+      });
+      this.mediaPaused = this.mediaSource.paused;
+    }
```
Allcharles, line 48: Just some general code cleanup which could be used through the project

```suggestion
    if (!this.mediaSource) {
      return
    }

    const updateMediaPausedState = () => this.mediaPaused = this.mediaSource.paused;
    this.mediaSource.addEventListener('pause', updateMediaPausedState);
    this.mediaSource.addEventListener('play', updateMediaPausedState);
    updateMediaPausedState();
```

src/components/media-controls/media-controls.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  // Action to be performed when the back button is pressed
+  @Prop() onBack: VoidFunction;
+
+  // Action to be performed when the forward button is pressed
+  @Prop() onForward: VoidFunction;
+
+  // Media source reference retrieved by getMedia
+  @State() mediaSource: HTMLMediaElement;
+  @State() mediaPaused: boolean;
```
Allcharles, line 27: Rename this

```suggestion
  @State() isMediaPaused: boolean;
```

Allcharles (approved):
I don't have a lot to add on top of Anthonys review. Mostly some of the variable names could be improved, and there is some repeated code that could be simplified.

PR #13, opened 2021-06-01
Remove global references in ewc-media-conrols, and associated helpers and tests

benjaminARowe:
### What
Remove global references from ewc-media-controls, and associated helpers and tests

### How
- A logger utility has been created, to prevent calls to console.log and console.warn
    - This logger is used in tests with mock functions to prevent writing to global console
- Use of HTMLElement and subclasses has been replaced with globalThis.HTMLElement
    - globalThis is used for accessing typing imported from typescript automatically (specifically typescript/lib/lib.dom.d)
- On keystroke linting has been fixed and enforces rules that prevent creation of variables in the global scope, overwriting variables in the global scope and use of some global variables

### Tests
- Many existing tests for ewc-media-controls have been temporarily removed. They will be added in a future pull request.


src/components/media-controls/media-controls.tsx
```diff
@@ -1,38 +1,42 @@
...
+
   // The id of the HTMLMediaElement
   @Prop() for: string;
 
-  // Media control strings
-  @Prop() back: string = 'Back';
-  @Prop() play: string = 'Play';
-  @Prop() pause: string = 'Pause';
-  @Prop() forward: string = 'Forward';
+  @Prop() logger;
```
atruskie, line 18: should not be a `@prop`

atruskie (changes requested):
Ok, I gave some misleading advice. 

It appears stencil has done a lot of work to make `window`, `document`, and such globals just work™.

What is clear is that a dependency like `logger` shouldn't be `@state`, or a `@prop`, or an attribute that can be bound to.

Here is my suggested layout:

```
// our own definitions which in their simplest form just refer to the originals
// but provides us with an opportunity to proxy/remove/change them later
import logger from ./logger;
import win, doc from ./globals;

class BaseComponent {
  // allows testing tools to sub in or mock these at runtime
  protected _logger = logger;
  protected _window = win;
  protected  _document = doc;
}
 
@Component({ ... })
class MediaControls extends BaseComponent {
 
}
```

Other notes:
- `globalThis` shouldn't appear in your code, use a reference to one of our globals above
- good job on the eslint config. Needs a little bit of tuning: standard patterns for this project should not have to be disabled everytime they're used.

src/components/media-controls/readme.md
```diff
@@ -7,11 +7,15 @@
...
-| `first`  | `first`   | The first name  | `string` | `undefined` |
-| `last`   | `last`    | The last name   | `string` | `undefined` |
-| `middle` | `middle`  | The middle name | `string` | `undefined` |
+| Property    | Attribute | Description | Type           | Default     |
+| ----------- | --------- | ----------- | -------------- | ----------- |
+| `back`      | `back`    |             | `string`       | `'Back'`    |
+| `for`       | `for`     |             | `string`       | `undefined` |
+| `forward`   | `forward` |             | `string`       | `'Forward'` |
+| `onBack`    | --        |             | `VoidFunction` | `undefined` |
+| `onForward` | --        |             | `VoidFunction` | `undefined` |
```
Allcharles, line 16: Fairly certain these were removed, update the properties. Potentially update git hooks to automatically update these files when you commit

src/utils/logger.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  // Produce a namespaced console warning
+  warn(message: any) {
+    if (this.console && this.console.warn != null) {
+      this.console.warn(`${this.name}: ${message}`);
+    }
+  }
+
+  // Produce a namespeced console log
```
Allcharles, line 19: Potentially install a spellchecker to catch these things
```suggestion
  // Produce a namespaced console log
```

src/utils/logger.ts
```diff
@@ -0,0 +1,25 @@
...
+    this.console = console ?? window.console;
+    this.name = name ?? 'ewc';
+  }
+
+  // Produce a namespaced console warning
+  warn(message: any) {
+    if (this.console && this.console.warn != null) {
+      this.console.warn(`${this.name}: ${message}`);
+    }
+  }
```
Allcharles, line 17: I'm not sure this will work with `any` input type. Potentially the following would be a better fit:

```suggestion
  warn(message: any) {
    if (typeof this.console?.warn !== "function") {
      // Handle this case somehow
    }
  
    if (["string", "number", "boolean"].includes(typeof message)) { 
      this.console.warn(`${this.name}: ${message}`);
    } else {
      this.console.warn(`${this.name}: `, message);
    }
  }
```

You could also extract this function further to reuse it between all of the different log types fairly easily. If you are going to build a logger, it should expose the same functionality as the Console (ie. `assert()`, `clear()`, `count()`, etc).

Allcharles (approved):
A couple of my comments from the previous PR apply to this one so I won't re-iterate them. Something i would like to see the use of to simplify your imports is [typescript paths](https://stackoverflow.com/questions/43281741/how-to-use-paths-in-tsconfig-json).

To add onto Anthony's review, I would recommend the following way of implementing the `BaseComponent` as it'll be more future proof and wont block you from extending multiple classes:

```typescript
// our own definitions which in their simplest form just refer to the originals
// but provides us with an opportunity to proxy/remove/change them later
import logger from ./logger;
import win, doc from ./globals;

// You may be able to restrict T to a more defined type, I'm unsure about StencilJS component types
export function withBaseComponent<T>(
  base: T = class {} as any
) {
  return class extends base {
    // allows testing tools to sub in or mock these at runtime
    protected _logger = logger;
    protected _window = win;
    protected  _document = doc;
  };
}

class CustomBaseComponent {
    public isCustomComponent = true;
}

@Component({ ... })
class MediaControls extends withBaseComponent() {
}

@Component({ ... })
class MediaControls extends withBaseComponent(CustomBaseComponent) {
}
```

Then for testing, you would create a function which allows you to individually override the `logger`, `window`, or `document` as you need. Such as:

/test/helpers.ts
```typescript
function overrideLogger(component, testLogger) {
    component["_logger"] = testLogger;
}
```

```typescript
beforeEach(() => {
    overrideLogger(component, testLogger);
});
```

PR #15, opened 2021-06-15
Add theming and remove globalthis use

benjaminARowe:
## What
- Add a global themeing file which contains variables that can be used in all other css files in the project to alter the theme
- Add a globals file that contains static references to win, doc and logger, which can be used in any file
## Why
- Global themeing will allow the theme for all components to be easily set
- Adding a single location where refereences to win, doc and logger are available will remove the need to access "globalThis" or overwrite other global variables
## How
- A global CSS file is included in the entrypoint for the component library
- A file that exports static references to document, window and logger was used. Creating a baseComponent which could be inherited from was preferred, however stencil.js does not allow this. This may be possible in the future with the "mixin" decorator, but the current proposal for these seems like it will be rejected.

## Notes
I'm looking for advice about:
- how to prevent components.d.ts from cluttering up the commits. It is automatically generated, however according to several issues, adding it to gitignore causes problems
- How to properly mock types for testing. See types.ts for my failed attempt at this. I need to figure out how to do this because the stencil.js version of jest seems to only contain empty shells of vanilla types, that do not inherit from eachother the same way as they usually do


src/test/types.ts
```diff
@@ -0,0 +1,5 @@
+export class TestElement extends Object {}
+
+export class TestHTMLElement extends TestElement {}
+
+export class TestHTMLMediaElement extends TestHTMLElement {}
```
atruskie, line 5: seems like a smell... let's see what you do with them below...

src/utils/globals.ts
```diff
@@ -0,0 +1,9 @@
+import logger from './logger';
+
+export default class Globals {
+  public static _logger = logger;
+
+  public static _win = window;
+
+  public static _doc = document;
```
atruskie, line 8: underscore prefix is unnecessary

atruskie (approved):
Okay the no mixins or inheritance is bonkers. Using a wrapper object (`Globals`) is sufficient in that if we do find a method to inject services/objects at runtime, we'll be able to break the `Globals` import reference and TS will tell us what breaks.

For example, it'd be fairly easy now to replace `Globals` with an object that can be configured. In testing we could do `Globals.configure(myCustomLogger)` and have `Globals.logger` return `myCustomLogger`. Don't make that change until needed... but at least with the layer of redirection you've added we have options down the line.

src/components/media-controls/media-controls.spec.tsx
```diff
@@ -1,23 +1,27 @@
...
-logger.warn = jest.fn();
-logger.log = jest.fn();
+beforeEach(() => {
+  let logger = globals._logger;
+  logger.log = jest.fn();
+  logger.warn = jest.fn();
+
+  //let win = globals._win;
+  globals._win.HTMLElement = TestHTMLElement;
+  globals._win.HTMLMediaElement = TestHTMLMediaElement;
```
atruskie, line 16: are you not actually modifying the original `window` reference here? Seems dangerous.

Also don't see how these are needed/used for this test.

atruskie:
@benjaminARowe 

>How to properly mock types for testing. See types.ts for my failed attempt at this. I need to figure out how to do this because the stencil.js version of jest seems to only contain empty shells of vanilla types, that do not inherit from eachother the same way as they usually do

Good question. I'd say, naively, we don't need to mock at all. 

Tests that involve rendering components should render in the browser. Unless there's a specific need to mock (to disable some complex / external behaviour) we should be rendering and testing real components. 

If that answer is not satisfactory, I'd need to play around in the testing environment to see what is what. Let's schedule a meeting.

Allcharles (commented):
PR looks good, in regards to the issue with `component.d.ts` it I would keep committing the file until the [issue](https://github.com/ionic-team/stencil/issues/1776) tracking it is reopened and then closed (or if they magically fix it one update). Otherwise there isn't much you can do unless they expose a command which generates the `component.d.ts` file before use.

PR #20, opened 2021-08-25
Ben/axes

### LINKED ISSUE #1: Axes Component
Shows axes and gridlines for a target

- Must have a target (`for`)
- Target may have coordinates
    - or else coordinates may be referred to
- Can enable or disable
    - x axis
    - y axis
    - x grid line
    - y grid lines
- Can show logarithmic axis
- Can customize
    - scale step
    - to show end tick (e.g. prioritize showing end value tick over last scale step tick)
    - axis labels
    - axis position

benjaminARowe:
## What
Add ewc-axes, ewc-x-axis, and ewc-y-axis. These components can be added to any HTMLElement to add a configurable set of axes. 

## Why
There is no built-in solution. A typical solution is to use d3 directly, but this has a relatively high barrier for entry. These components will allow axes to be added to any element with a minimum of just specifying the for, min and max properties.

## How
ewc-axes the id of the target element is passed to the "for" property. ewc-axes then uses the ResizeObservable API and rxjs to measure and wrap itself around the target element. ewc-axes contains an svg to serve as a rendering surface. ewc-x-axis and ewc-y-axis can be added as children, and draw axes to the svg contained by ewc-axes.

## Testing
Testing for this component has not yet been implemented due to issues with inheritance in the testing environment.

## TODO (To be added as issues)
- Allow getting min and max from a HTMLMediaElement
- Allow "for" property to also accept a reference

Closes #1


src/utils/axes.ts
```diff
@@ -0,0 +1,26 @@
+import { axisLeft, axisBottom } from 'd3-axis';
+import { select } from 'd3-selection';
+import { scaleLinear } from 'd3-scale';
+
+export function drawYAxis(surface, size, border, dataMin, dataMax, ticks) {
+  let scale = scaleLinear().domain([dataMin, dataMax]).range([0, size.height]);
+  let translate = `translate(${border}, 0)`;
+  drawAxis(surface, size, axisLeft, scale, translate, 'y_axis', ticks);
+}
```
Allcharles, line 9: Types? And use const instead of let

src/utils/axes.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export function drawAxis(surface, size, axis, scale, translate, id, ticks) {
+  let width = size.width,
+    height = size.height;
+
+  let svg = select(surface).attr('width', width).attr('height', height);
+
+  let a = axis().scale(scale);
+  ticks != null ? a.ticks(ticks, 'f') : null;
```
Allcharles, line 23: What if ticks = undefined or 0?

src/utils/axes.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  svg.select(`#${id}`).node() ? svg.select(`#${id}`).call(a) : svg.append('g').attr('transform', translate).attr('id', id).call(a);
```
Allcharles, line 25: Turn this into an if statement for clarity. Ternary statements are mainly for returning values

src/utils/observables.ts
```diff
@@ -0,0 +1,15 @@
+import * as rxjs from 'rxjs';
```
Allcharles, line 1: Don't use * imports, this makes it harder for tree shaking

src/utils/observables.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export function resizeObservable(elem) {
```
Allcharles, line 3: Type inputs

Allcharles (changes requested):
I've left a lot of comments on things I thought needed changing. Some of them may be outside of scope, and I'll leave it to Anthony and yourself to decide if that is the case. I also pulled you up on a couple things which may be bugs down the road.

PR #28, opened 2024-04-05
Spectrogram tests

hudson-newey:
- Adds spectrogram tests & fixtures
- Gets functionality working for the `test/helper` functions (these functions still have sub-optimal TS typing)
- Duck types the spectrogram component with the same events, properties, and attributes as the `HTMLAudioElement` (this allows `oe-media-controls` to access both elements through the same APIs and event bindings)
	- Changes the name of the `playing` property to `paused`
	- Changes the name of the `playing` event to `play`

I've done this to duck type the spectrogram component with `HTMLAudioElement` events, properties and attributes. This allows `oe-media-controls` to control spectrogram components and `HTMLAudioElement`'s through the same API's and events

## Added tests

- `mount` tests that assert the correct events are emitted when `play()` and `pause()` are called
- tests to check that a `<source>` element can be used as a media source in the spectrogram components default `<slot>`
- test to check that `oe-media-controls` can correctly mount/unmount and re-mount to spectrograms correctly
- Remove the `oe-media-controls` unit test example that is no longer needed (we originally created this as a proof of concept on how to test capturing events. Additionally, `playing` state is now handled by the `oe-spectrogram` component) this test is no longer needed as we have a practical example of testing captured events in the `oe-spectrogram` component


atruskie:
> I have purposely not used conventional commits so that the package version isn't bumped and a release isn't created for these WIP components.

You should be even now. The whole history should be conventional, that's the point. We should disable the publish step instead.

PR #50, opened 2024-05-16
Change axes component to render in Lit

hudson-newey:
# Change axes component to render in Lit

By opening this pull request, we will get a preview build of the documentation site. We can use the preview build to demo examples and use cases.

## Changes

**Features**

- Adds support for rendering spectrograms in a Lit component
- Adds support for reactively changing the source of a spectrogram

**Bug Fixes**

- Fixes bug where changing the window size would not paint to the correct canvas size
- Fix various bugs to do with resizing the spectrogram
- Fix bug where production docs deployment wouldn't use the `netlify.toml` (this is fixed by having a docs-publish workflow that takes a "production" parameter)
- Fix race condition where changing spectrogram options (e.g. colour) can cause the spectrogram to partially render

Race conditions fixed in: #51 

**Code Cleanup**

- Removes D3 dependencies (apart from d3-scale)
- Refactor `audio.ts` so that common utilities between re-rendering and initialization use the same methods
- Cache HTTP responses and metadata between re-painting
- Cache the UnitConverts for the indicator component

## Preview

N.a.


src/components/spectrogram/spectrogram.ts
```diff
@@ -114,33 +114,41 @@ export class Spectrogram extends SignalWatcher(AbstractComponent(LitElement)) {
...
+    if (this.doneFirstRender) {
+      // spectrogram regeneration functionality
+      if (this.invalidateSpectrogramOptions(change)) {
+        this.audioHelper.regenerateSpectrogramOptions(this.spectrogramOptions());
+      }
+
+      if (this.invalidateSpectrogramSource(change)) {
+        this.currentTime.value = 0;
+        this.regenerateSpectrogram();
+      }
```
atruskie, line 131: probably want to make sure we dont do both (else if)

src/helpers/audio/audio.ts
```diff
@@ -0,0 +1,211 @@
...
+    this.spectrogramWorker = new WorkerConstructor();
+  }
+
+  private readonly spectrogramWorker: Worker | undefined;
+  private processorNode: AudioWorkletNode | undefined;
+  private state: State | undefined;
+  private sampleBuffer: SharedArrayBuffer | undefined;
+  private offscreenCanvas: OffscreenCanvas | undefined;
+  private cachedFile!: IFetchedAudio<Response>;
+  private segmentSize = 41100;
```
atruskie, line 25:
```suggestion
  private segmentSize = 44100;
```

src/helpers/audio/audio.ts
```diff
@@ -0,0 +1,211 @@
...
+
+    // the number of samples after which to trigger a render of the spectrogram
+    // the balance of this number is a performance tradeoff
+    // - too many samples and we'll use more memory and render in larger clunky chunks
+    // - too few samples and we'll be rendering too often and hit performance bottlenecks
+    //   e.g. with canvas painting, wasm interop, signalling primitives, etc.
+    //
+    // about one seconds-worth of samples
+    this.state = State.createState(this.segmentSize);
+    this.sampleBuffer = new SharedArrayBuffer(Float32Array.BYTES_PER_ELEMENT * this.segmentSize);
```
atruskie, line 44: this can move to the constructor now

src/helpers/audio/audio.ts
```diff
@@ -0,0 +1,211 @@
...
+
+    if (!this.canvasTransferred) {
+      this.offscreenCanvas = canvas.transferControlToOffscreen();
+    }
+
+    this.setupWorker(metadata, options);
+    this.setupProcessor(context);
+
+    console.time("rendering");
```
atruskie, line 53: wrong spot?

src/components/spectrogram/spectrogram.ts
hudson-newey: Yeah, I found that this caused a bug with re-rendering src attributes and have fixed it in an upcoming commit

src/helpers/audio/audio.ts
hudson-newey: Moved to the `setupProcessor` function above `startRendering()`

Issue #68, opened 2024-07-09
Classification use cases


hudson-newey:
Users have expressed the desire to create datasets using the verification interface

From this additional use cases have arised

|     | Classifying/Verifying | Multi Class | Mixed Dataset | Additional Tags? | Persona Use Case |
| --- | --------------------- | ----------- | ------------- | ---------------- | ---------------- |
| 1   | Classifying           | TRUE        | N/A           | N/A              | Phil             |
| 2   | Classifying           | FALSE       | N/A           | N/A              | Slade            |
| 3   | Verifying             | N/A         | FALSE         | FALSE            | Slade            |
| 4   | Verifying             | N/A         | TRUE          | FALSE            | NSW/BirdNet      |
| 5   | Verifying             | N/A         | FALSE         | TRUE             | Dani             |
| 6   | Verifying             | N/A         | TRUE          | TRUE             | Unknown??        |

To differentiate between classification and verification, the `oe-decision` element can be bifurcated into `oe-classification` and `oe-verification`.

Because classification can involve adding multiple classifications to one recording, auto-paging will have to be disabled, users should be able to add multiple decisions to one recording snippet, and they should also be able to remove decisions (e.g. through right clicking a tile)

### Ideas

During verification tasks, the tag is already known, therefore, the `tag`
attribute can be omitted on `oe-decision` elements.

During classification tasks, the user is applying additional tags to a subject.
Therefore, the `tag` attribute must be emitted, while the `verified` attribute
must be omitted.

PR #104, opened 2024-07-30
New verification/classification mental models

hudson-newey:
# New verification/classification mental models

### Features

- Adds support for classification use cases
- Refactor the `getPage` to support iterator functions through context objects
- Adds tag recognition for "scientificName" and "commonName" column names
- Adds support for space separated column keys such as "common name"

### Bug Fixes

- Fixed bug where the info card would not display the "show more" button when subjects had more than 3 key-value pairs
- Makes the play/pause buttons $1.5 \times$ larger (so that it appears to be the same size as the rest of the media control icons)
- Fixes bug where colorBrewer was using a linear gradient (not categorical)
- Fixes bug where multiple additional tags would not be separated by a comma

### Code Quality

- Add internal documentation about documented use cases
- Now uses the theme colors for the the verification grid tile hover effect (was previously a constant)
	- I also reduced the `--oe-backdrop-shadow` by 33% (this was only ever used in the grid tile component)
- `queryAllDeeplyAssignedElements` now returns an empty array if no elements match the selector
- Moved `Annotation`, `Decision`, `Verification`, and `Classification` class models into their own separate files (they were previously all in the one file)
- Generated [snapshot tests](https://playwright.dev/docs/test-snapshots)
- Runs prettier over the code base to fix inconsistent indentation

## Visual Changes

image

_New classification use cases_

---

image
image

_Old & new icon sizes for the media controls play/pause button_

---

image
image

_Old & new drop shadow for hovering over a tile_

## Related Issues

Fixes: #68 

## Final Checklist


src/components/verification-grid/verification-grid.ts
```diff
@@ -893,49 +866,161 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+        if (!exclusiveTag.decisionModel) {
+          continue;
+        }
+
+        tile.removeDecision(exclusiveTag.decisionModel);
+      }
     }
 
-    this.decisionsDisabled = disabled;
+    tile.addDecision(decision);
```
atruskie, line 893: this logic feels out of place. I think some of it belongs on the subject wrapper. I think the only part of this that should remain in here is `tile.addDecision`. Modifying the state of the subject wrapper should be the responsibility of the subject wrapper. 

I also think you can simplify the replacement logic if you used a dictionary/map on the subject wrapper to store decisions. e.g. `subject.deicsions[verificationSymbol] = decision`. You could split storage of classifications/verifications on the subject wrapper too.

I don't understand the `exclusiveToTags` part of the code. Is this a toggle implementation?

src/components/verification-grid/verification-grid.ts
```diff
@@ -893,49 +866,161 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
-  //#endregion
+  private removeTileDecision(tile: VerificationGridTileComponent, decision: IDecision): void {
+    if (tile.model.decisions.length === 0) {
+      throw new Error("Cannot remove decision from tile that has no decisions");
+    }
 
-  //#region DecisionHighlights
+    tile.removeDecision(decision);
+    this.subjectHistory = this.subjectHistory.filter((model) => model.identifier !== decision.identifier);
+  }
```
atruskie, line 903: same criticism as a above

src/components/verification-grid/verification-grid.ts
```diff
@@ -893,49 +866,161 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
-      const gridTile = this.gridTiles[i];
-      gridTile.decisionIndex = decision.origin;
-    });
+  // TODO: finish this function
+  private shouldAutoPage(): boolean {
+    // we should auto page if there are no more decisions that can be made
+    // for classification tasks, this means that each tile has a positive or
+    // negative decision made about every tag
+    // during verification tasks, it means that each tile either has a positive
+    // negative, skip, or unsure decision applied to it
```
atruskie, line 917: conflicting statements

src/components/verification-grid/verification-grid.ts
atruskie, line 1: you're using tags a lot in place of `decision` in this document. e.g. `outstandingTags`. I think they're all decisions, no? tighten up the language

src/components/verification-grid/verification-grid.ts
```diff
@@ -893,49 +866,161 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+      const tileTags = tile.model.classifications.map((classification) => classification.tag.text);
+      const tileHasRequiredTags = requiredTags.every((tag) => tileTags.includes(tag.text));
+
+      if (!tileHasRequiredTags) {
+        return true;
+      }
+    }
+
+    return false;
+  }
```
atruskie, line 995: ok i have a gut feeling all of this could be reduced to about 5 lines of code. Need to have a think about this.

If we could ask each subject what decisions it has (e.g. `tile.subject.hasDecisions(allRequiredDecisions)` this is just a forr loop, no?

I also don't understand the `outstandingTags` function.

src/components/verification-grid/verification-grid.ts
```diff
@@ -893,49 +866,161 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+
+  private setDecisionDisabled(disabled: boolean): void {
```
atruskie, line 997:
```suggestion
  private setDecisionsDisabled(disabled: boolean): void {
```

src/components/verification-grid/verification-grid.ts
```diff
@@ -1081,25 +1165,82 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+
+  private mixedTaskPromptTemplate(hasMultipleTiles: boolean, hasSubSelection: boolean): TemplateResult<1> {
+    if (hasSubSelection) {
+      return html`<p>Make a decision about all of the selected</p>`;
+    }
+
+    if (hasMultipleTiles) {
+      return html`<p>Make a decision about all of the tiles</p>`;
+    } else {
+      return html`<p>Make a decision about the shown spectrogram</p>`;
```
atruskie, line 1186: mixed vernacular, and a third term, _subject_ in the section below

src/components/verification-grid/verification-grid.ts
```diff
@@ -1081,25 +1165,82 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+
+  private classificationTaskPromptTemplate(hasMultipleTiles: boolean, hasSubSelection: boolean): TemplateResult<1> {
+    if (hasSubSelection) {
+      return html`<p>Apply labels to selected subjects</p>`;
+    }
+
+    if (hasMultipleTiles) {
+      return html`<p>Apply all relevant labels to all relevant subjects</p>`;
+    } else {
+      return html`<p>Apply all relevant labels to the subject</p>`;
```
atruskie, line 1198: even more mixed vernacular: you use _classification_ in the help prompt so you should use _classify_ here

src/components/verification-grid/verification-grid.ts
```diff
@@ -1081,25 +1165,82 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+
+  private verificationTaskPromptTemplate(hasMultipleTiles: boolean, hasSubSelection: boolean): TemplateResult<1> {
+    if (hasSubSelection) {
+      return html`<p>Do all of the selected have the correct applied tag</p>`;
+    }
+
+    if (hasMultipleTiles) {
+      return html`<p>Do all of the tiles have the correct applied tag</p>`;
+    } else {
+      return html`<p>Does the show spectrogram have the correct applied tag</p>`;
```
atruskie, line 1210: ...

src/components/verification-grid/verification-grid.ts
```diff
@@ -1123,11 +1264,68 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
     `;
   }
 
+  // TODO: this function could definitely be refactored
```
atruskie, line 1267: agreed. More of this should be in another place.

Ideally: `tile.subject.skipUndecided(requiredDecisions)`

src/helpers/attributes.ts
```diff
@@ -47,3 +48,11 @@ export const callbackConverter = (value: string | ((...params: any) => any)) =>
 
   return value;
 };
+
+export const enumConverter = <T extends Enum>(enumValues: T) => {
```
atruskie, line 52: is this case insensitive? it should be

src/models/decisions/classification.ts
```diff
@@ -0,0 +1,26 @@
...
+    origin: number;
+}
+
+export class Classification implements IClassification, IDecision {
+    public constructor(data: IClassification) {
+        this.tag = data.tag;
+        this.confirmed = data.confirmed;
+        this.origin = data.origin;
+
+        this.identifier = Symbol("classification");
```
atruskie, line 18: this means no classification is ever equal to any other classification?

src/models/decisions/decision.ts
```diff
@@ -0,0 +1,11 @@
+import { EnumValue } from "../../helpers/advancedTypes";
+import { VerificationDecision } from "./verification";
+
+// A polymorphic structure that can be used to represent any decision type
+// that implements the interface
+export interface IDecision {
+    confirmed: EnumValue<VerificationDecision>;
+    identifier: symbol;
+    type: string;
+    origin: number;
```
atruskie, line 10: what does origin mean?

src/models/decisions/verification.ts
```diff
@@ -0,0 +1,44 @@
...
+    this.origin = data.origin;
+
+    this.identifier = Symbol("verification");
+  }
+
+  public type = "verification";
+  public identifier: symbol;
+  public origin: number;
+  public confirmed: EnumValue<VerificationDecision>;
+  public additionalTags: Tag[];
```
atruskie, line 43: should not longer have additional tags

src/models/subject.ts
```diff
@@ -0,0 +1,50 @@
...
+
+// the Verification model is emitted by the oe-verification-grid as a CustomEvent
+export class SubjectWrapper implements ISubjectWrapper {
+    public constructor(data: ISubjectWrapper) {
+        this.subject = data.subject;
+        this.decisions = data.decisions ?? [];
+        this.url = data.url;
+        this.tag = data.tag;
+
+        this.identifier = Symbol("decision");
```
atruskie, line 26: inappropriate symbol name

src/models/subject.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    // aka: context
+    // this is the native data model used by the host application
+    // or this could be the csv row
```
atruskie, line 31: misplaced comment

src/models/subject.ts
```diff
@@ -0,0 +1,50 @@
...
+    public identifier: symbol;
+    public subject: VerificationSubject;
```
atruskie, line 33: why are we double wrapping the subject? A classification has a subject too, but the type name is `VerificationSubject`??

src/models/subject.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    // each row can only have one verification decision
+    // but can have multiple classification decisions
+    // verification decisions will be reflected in the oe-confirmed
+    // column, while each classification will get its own row
+    public decisions: [Verification, ...Classification[]] | Classification[];
```
atruskie, line 39: based on what else i've seen in this review, this array should be a getter and you should store Verifications and classifications separately (or as a map)

src/services/gridPageFetcher.ts
atruskie, line 1: it occurs to me that it would far simpler to track the cache state if we attached that information to the subject wrapper.

src/services/gridPageFetcher.ts
```diff
@@ -1,116 +1,114 @@
...
 
-export type PageFetcher = (elapsedItems: number) => Promise<DecisionWrapper[]>;
+/**
+ * A context object that is passed to the page fetcher function after every call
+ * this can be used to keep track of state such as the current page, where the
+ * page fetcher would increment the page number after every call.
+ */
+export type PageFetcherContext = Record<string, unknown>;
+export type PageFetcher<T extends PageFetcherContext = any> = (context: T) =>
+    Promise<SubjectWrapper[]>;
```
atruskie, line 11: we should allow for immutable context. We store their last context and pass it back to them on next invocation.

```suggestion
    Promise<{ subjects: SubjectWrapper[], context: T}>;
```

atruskie (changes requested):
Great work. I reckon a big focus is needed on working out who should be responsible for bits of state.

hudson-newey:
Additional comments from user testing:

- The "Continue Verifying" should be "Continue Classifying" during a classification only task

src/components/verification-grid-tile/verification-grid-tile.ts
```diff
@@ -219,11 +249,53 @@ export class VerificationGridTileComponent extends AbstractComponent(LitElement)
...
+        const last = index === decisionIndices.length - 1;
+
+        const percentage = percentageStep * (index + 1);
+        if (first) {
+          return `var(--decision-color-${color}) ${percentage}%`;
+        } else if (last) {
+          return `var(--decision-color-${color}) 0`;
+        }
+
+        return `var(--decision-color-${color}) 0 ${percentage}%`
```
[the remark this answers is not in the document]
hudson-newey: Yes, we can simplify quite a bit

image

src/components/verification-grid/verification-grid.ts
```diff
@@ -97,22 +105,23 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
   @queryAllDeeplyAssignedElements({ selector: "oe-decision" })
-  private decisionElements?: DecisionComponent[];
+  private oeDecisionElements!: DecisionComponent[];
+
+  /** selector for oe-verification elements */
+  @queryAllDeeplyAssignedElements({ selector: "oe-verification" })
+  private verificationDecisionElements!: VerificationComponent[];
+
+  /** selector for oe-classification elements */
+  @queryAllDeeplyAssignedElements({ selector: "oe-classification" })
```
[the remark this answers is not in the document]
hudson-newey: This was a conscious design decision because slots are user defined.

In my opinion, users should have the ability to (but shouldn't) create a document template like

```html
<oe-verification-grid>
	<template>
		<!-- snipped -->
	</template>

	<nav class="sidebar">
		<oe-classification>Classification in side bar</oe-classification>
	</nav>
</oe-verification-grid>
```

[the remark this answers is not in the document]
atruskie: Not that I want the use of `slot` encouraged here either, but isn't that the role that `slot` should be filling?

```
<oe-verification-grid>
	<template>
		<!-- snipped -->
	</template>

	<nav slot="classification" class="sidebar">
		<oe-classification>
			<p>Classification in side bar</p>
			<p>That is also not in the default slots top level</p>
		</oe-classification>
	</nav>
</oe-verification-grid>
```

Regardless, I don't think we need this level of customization. If the `<oe-classification/verification>` element is not the first child it is configured incorrectly.

If we ever need more customization here - which seems unlikely given the constraints of the UI - we make this more flexible later.

src/components/verification-grid/verification-grid.ts
```diff
@@ -710,12 +738,13 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
-    const decisionEnd = Math.min(this.decisions.length, decisionStart + this.gridSize);
-    const decisionHistory = this.decisions.slice(decisionStart, decisionEnd);
+    const decisionStart = Math.max(0, this.subjectHistory.length - historyOffset);
+    const decisionEnd = Math.min(this.subjectHistory.length, decisionStart + this.gridSize);
+    const decisionHistory = this.subjectHistory.slice(decisionStart, decisionEnd);
 
-    this.historyMode(decisionHistory);
     this.renderVirtualPage(decisionHistory);
+    this.removeDecisionButtonHighlight();
+    this.historyMode();
```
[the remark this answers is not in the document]
hudson-newey: Related issue: #139

src/models/decisions/classification.ts
hudson-newey: Kind of.

I use this symbol to check if a decision has been applied to a tile or not

src/components/verification-grid-tile/verification-grid-tile.ts
```diff
@@ -137,6 +145,22 @@ export class VerificationGridTileComponent extends AbstractComponent(LitElement)
...
   }
 
+  public addDecision(decision: IDecision) {
+    // because the model is an object (not a primitive), modifying a property
+    // does not cause the re-render which is needed to display the new decision
+    // as a border color
+    // to fix this, we call requestUpdate which will re-render the component
+    // TODO: We can probably replace this with a guard directive
+    this.model.addDecision(decision);
+    console.log(decision);
```
atruskie, line 153: unscoped log. should be using our log service

src/components/verification-grid/verification-grid.ts
```diff
@@ -71,15 +74,18 @@ interface HighlightSelection {
  * @slot - A template element that will be used to create each grid tile
  * @slot - Decision elements that will be used to create the decision buttons
  * @slot data-source - An `oe-data-source` element that provides the data
+ * @slot classifications - Grouped `oe-classification` components
+ * @slot verifications - Grouped `oe-verification` components
```
atruskie, line 78: not part of the spec? incorrect documentation?

src/components/verification-grid/verification-grid.ts
```diff
@@ -71,15 +74,18 @@ interface HighlightSelection {
...
  * @fires loaded - Emits when all the spectrograms have been loaded
  */
 @customElement("oe-verification-grid")
 export class VerificationGridComponent extends AbstractComponent(LitElement) {
   public static styles = unsafeCSS(verificationGridStyles);
-  private static loadedEventName = "loaded" as const;
-  private static decisionMadeEventName = "decision-made" as const;
+
+  public static readonly decisionMadeEventName = "decision-made" as const;
+  private static readonly loadedEventName = "loaded" as const;
```
atruskie, line 88: maybe future issue: our events should be namespaced

src/components/verification-grid/verification-grid.ts
```diff
@@ -304,6 +326,15 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
       return;
     }
 
+    if (event.ctrlKey && event.key === "d") {
```
atruskie, line 329: did we add the appropriate shortcut hint in the bootstrap modal?

src/components/verification-grid/verification-grid.ts
```diff
@@ -304,6 +326,15 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+      event.preventDefault();
+      this.removeSubSelection();
+
+      if (!this.isViewingHistory()) {
+        this.removeDecisionButtonHighlight();
+      }
```
atruskie, line 335: this leaves me with so many questions - but knowing we're planning on changing this i'm just going to let this slide

src/helpers/attributes.ts
```diff
@@ -47,3 +48,18 @@ export const callbackConverter = (value: string | ((...params: any) => any)) =>
...
+  return (value: string): T[keyof T] | undefined => {
+    // we compare the requested key, and the enums keys as lowercase so that
+    // the attribute that is exposed to the user is case insensitive
+    const lowercaseKey = value.toLowerCase();
+    const enumKey = Object.keys(enumValues).find(
+      (key) => key.toLowerCase() === lowercaseKey,
```
atruskie, line 58: Not that it matters in this case, but I'd generally recommend `localeCompare`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#comparing_strings

src/models/decisions/classification.ts
```diff
@@ -0,0 +1,31 @@
+import { EnumValue } from "../../helpers/advancedTypes";
+import { Tag } from "../subject";
+import { IDecision, VerificationDecision } from "./decision";
+
+export interface IClassification {
+    tag: Tag;
+    confirmed: EnumValue<VerificationDecision>;
+    colorIndex: number;
+    identifier?: symbol;
+}
```
atruskie, line 10: redundant interface?

src/models/decisions/decision.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+// A polymorphic structure that can be used to represent any decision type
+// that implements the interface
+export interface IDecision {
+    confirmed: EnumValue<VerificationDecision>;
+    identifier: symbol;
+    type: string;
+    colorIndex: number;
+}
```
atruskie, line 17: Seems like an abstract class would have offered the same flexibility with less verbosity.

src/models/decisions/decision.ts
```diff
@@ -0,0 +1,17 @@
...
+export enum VerificationDecision {
+    FALSE = "false",
+    TRUE = "true",
+    UNSURE = "unsure",
+    SKIP = "skip",
+}
+
+// A polymorphic structure that can be used to represent any decision type
+// that implements the interface
+export interface IDecision {
```
atruskie, line 12: This interface has overlapping fields with IClassification and IVerification. Why?

src/models/decisions/decision.ts
```diff
@@ -0,0 +1,17 @@
...
+    confirmed: EnumValue<VerificationDecision>;
+    identifier: symbol;
+    type: string;
+    colorIndex: number;
```
atruskie, line 16: Why does this model know anything about colors?

Suggestion: store a decision identifier / key. Could be a type alias for number: `type DecisionId = number`. Or could be a decision source (I believe this what you intended `origin` to mean in the previous patch).

src/models/decisions/decision.ts
```diff
@@ -0,0 +1,17 @@
...
+    UNSURE = "unsure",
+    SKIP = "skip",
+}
+
+// A polymorphic structure that can be used to represent any decision type
+// that implements the interface
+export interface IDecision {
+    confirmed: EnumValue<VerificationDecision>;
+    identifier: symbol;
+    type: string;
```
atruskie, line 15: what is type?

src/models/decisions/decision.ts
```diff
@@ -0,0 +1,17 @@
...
+    TRUE = "true",
+    UNSURE = "unsure",
+    SKIP = "skip",
+}
+
+// A polymorphic structure that can be used to represent any decision type
+// that implements the interface
+export interface IDecision {
+    confirmed: EnumValue<VerificationDecision>;
+    identifier: symbol;
```
atruskie, line 14: what is this identifying?

src/models/decisions/verification.ts
```diff
@@ -0,0 +1,22 @@
...
+    this.colorIndex = data.colorIndex;
+
+    this.identifier = data.identifier ?? Symbol("verification");
+  }
+
+  public type = "verification";
+  public identifier: symbol;
+  public colorIndex: number;
+  public confirmed: EnumValue<VerificationDecision>;
+}
```
atruskie, line 22: why is there no tag field here?

src/models/decisions/decision.ts
```diff
@@ -0,0 +1,17 @@
...
+    FALSE = "false",
+    TRUE = "true",
+    UNSURE = "unsure",
+    SKIP = "skip",
+}
+
+// A polymorphic structure that can be used to represent any decision type
+// that implements the interface
+export interface IDecision {
+    confirmed: EnumValue<VerificationDecision>;
```
atruskie, line 13: why enumvalue and not just the enum?

src/helpers/advancedTypes.ts
```diff
@@ -0,0 +1,15 @@
+import { PropertyDeclaration } from "lit";
+
+/** Can be used to type a TypeScript enum */
+export type Enum = Readonly<{ [key: number]: string | number }>;
+
+/** A wildcard value from an enum */
```
atruskie, line 6: - wildcard means a value that matches any value, it does not mean choosing one value from the set

src/models/decisions/decision.ts
```diff
@@ -0,0 +1,17 @@
+import { EnumValue } from "../../helpers/advancedTypes";
+
+export enum VerificationDecision {
+    FALSE = "false",
+    TRUE = "true",
+    UNSURE = "unsure",
+    SKIP = "skip",
```
atruskie, line 7: why constant case here?

src/models/subject.ts
```diff
@@ -0,0 +1,157 @@
...
+
+export class Subject {
+    public constructor(data: SubjectData) {
+        this.data = data;
+        this.identifier = Symbol("verification-subject");
+    }
+
+    public identifier: symbol;
+    public data: SubjectData;
+}
```
atruskie, line 32: I ignored this last time, but it seems a lot to create a wrapper object for another object just to stick an unique identifier on an object in a dynamic language.

```
interface ISubject  = unknown & { identifier: Symbol }
```

Also document the purpose of the identifier symbol (equality i believe?)

src/models/subject.ts
```diff
@@ -0,0 +1,157 @@
...
+export interface ISubjectWrapper {
+    subject: Subject;
+    identifier: symbol;
+    clientCached: boolean;
+    serverCached: boolean;
+
+    url: string;
+    decisions: Map<symbol, IDecision[]>;
+    tag: string;
+}
```
atruskie, line 16: does this interface serve a purpose?

src/models/subject.ts
```diff
@@ -0,0 +1,157 @@
...
+        this.url = data.url;
+        this.tag = data.tag;
+
+        this.identifier = Symbol("decision");
+    }
+
+    // aka: context
+    // this is the native data model used by the host application
+    // or this could be the csv row
+    public identifier: symbol;
```
atruskie: hangon, didn't subject have an identifier?

why does the wrapper need an identifier? what purpose does it serve?

src/models/subject.ts
```diff
@@ -0,0 +1,157 @@
...
+    public decisions: Map<symbol, IDecision[]>;
+    public url: string;
+    public tag: string;
+
+    public get decisionModels(): IDecision[] {
+        return Array.from(this.decisions.values()).flat();
+    }
+
+    public get verification(): Verification | undefined {
+        return this.decisionModels.find((decision) => decision.type === "verification") as Verification;
```
atruskie, line 67: hang on, you added a map! this should be an O(1) lookup!?

src/models/subject.ts
```diff
@@ -0,0 +1,157 @@
...
+    public subject: Subject;
+
+    public clientCached = false;
+    public serverCached = false;
+
+    // each row can only have one verification decision
+    // but can have multiple classification decisions
+    // verification decisions will be reflected in the oe-confirmed
+    // column, while each classification will get its own row
+    public decisions: Map<symbol, IDecision[]>;
```
atruskie, line 58: It's not clear what the symbol is identifying here. A decision button? or a decision model class?

I believe it's the button component because you're grouping these things together.

Have you considered not grouping them?

src/models/subject.ts
```diff
@@ -0,0 +1,157 @@
...
+        return this.decisionModels.filter((decision) => decision.type === "classification") as Classification[];
+    }
+
+    public addDecision(newDecision: IDecision): void {
+        const decision = newDecision;
+
+        // if there are two decisions that are mutually exclusive, we want to
+        // remove them before applying the new decision
+        if (decision instanceof Verification) {
+            const existingVerification = this.verification;
```
atruskie, line 80: ok... this is a O(n) lookup in a common action, and not an occasional query method. We can do better.

src/models/subject.ts
```diff
@@ -0,0 +1,157 @@
...
+            if (existingVerification) {
+                this.removeDecision(existingVerification);
+            }
```
atruskie, line 83: none of this should be necessary.

```
this.decisions.set(validationKey, decision)
```

would overwrite the value. so why remove it first?

src/models/subject.ts
```diff
@@ -0,0 +1,157 @@
...
+
+        if (this.decisions.has(decision.identifier)) {
+            // TODO: remove this once TS can recognize that we have check the
+            // key exists in the guard that uses Map.prototype.has
+            const currentValue = this.decisions.get(decision.identifier) as IDecision[];
+            this.decisions.set(decision.identifier, [...currentValue, decision]);
+        }
+
+        this.decisions.set(decision.identifier, [decision]);
+    }
```
atruskie, line 102: ok I can see you've moved the logic to the model, so one point. Model knows how to apply decisions.

But there's still so much logic here for what should be a simple map set/get. Are you sure the structure is right?

src/models/subject.ts
```diff
@@ -0,0 +1,157 @@
...
+            ),
+        );
+        for (const tag of missingTags) {
+            const skipClassification = new Classification({
+                confirmed: VerificationDecision.SKIP,
+                colorIndex: 0,
+                tag,
+            });
+            this.addDecision(skipClassification);
+        }
```
atruskie, line 133: you've got a double loop here where one would  do.

atruskie (changes requested):
It looks like some kind of lint rule is inlining decorators. Reverse that change.

public/readme.txt
```diff
@@ -0,0 +1,4 @@
+ecosounds.csv is taken from the Ecoounds project "ACT Gov Bioacoustic Monitoring"
+under the Creative Commons by Attribution 4.0 license
```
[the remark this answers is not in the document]
hudson-newey: Yeah, I realised this and I changed to the "Cooloola" project which is CC

https://github.com/ecoacoustics/web-components/blob/classification/public/readme.txt

src/components/verification-grid/verification-grid.ts
hudson-newey: Related issue: https://github.com/ecoacoustics/web-components/issues/142

src/components/verification-grid/verification-grid.ts
hudson-newey: Good catch. I no longer need this (and it may have even caused a bug)

This was originally in here so that when you pressed escape to cancel a decision, the decision button highlight was removed.

This made sense when we could only apply one decision per tile, and we only showed touched decision buttons when viewing history.

src/models/decisions/decision.ts
hudson-newey: Type is no longer needed

I used to use it to identify if a decision was a `Verification` or `Classification` (Classification used to extend Verification so I could not use `instanceof`)

src/models/decisions/classification.ts
hudson-newey: Removed interface and changed constructor to use constructor argument injection instead of the anonymous object pattern

src/models/decisions/decision.ts
hudson-newey: I have removed these interfaces as per our offline discussion

Using an abstract class was also a good idea as it removes the overlapping fields

src/models/decisions/verification.ts
hudson-newey: As per our offline discussion, I've added a `tag` to the abstract `Decision` class (which both `Verification` and `Classification` extend)

This also has the added benefit of simplifying the `addDecision` method on the `Subjects` model, as we no longer have to remove decisions with the same tag

image

src/helpers/advancedTypes.ts
hudson-newey: New proposed JSDoc comment

image

src/models/subject.ts
hudson-newey: Removed as per our offline discussion

src/models/subject.ts
hudson-newey: As per offline discussions, the decisions are now indexed by tag

`Map<Tag, Decision>`

src/models/subject.ts
hudson-newey: Replaced with a one-line solution 🚀

image

src/models/subject.ts
hudson-newey: `SubjectWrapper` used to be `DecisionWrapper` a few commits back and was used to add/remove decisions

The identifier is no longer needed

src/models/decisions/decision.ts
hudson-newey: Strangely, it identifies where a decision came from (e.g. what decision button the decision model was created in)

Since we're leaning into `decisionId` (`origin`) uniquely identifying where a decision came from, I'm thinking that I'll remove `identifier` and use `decisionId` in its place

src/models/subject.ts
hudson-newey: I originally implemented this change like

image

But I'm thinking of reverting the change because `SubjectData` is user defined.
This means that if we add our identifier to the original row, the user cannot have a column named `identifier`.

We could namespace it (or other solutions) to reduce the risk of key collisions.
But I think adding a symbol to the user's input data row goes against our "we do not modify the original data" pitch

hudson-newey: Discussed offline:

Using a symbol as a key on the `subjectData` solves these concerns

atruskie (approved):
We've talked about a few more inconsistencies and simplifications for the classification/verification models offline. Once those are done and the tests are good, this is ready to merge

Issue #130, opened 2024-08-02
Prompt users to enter the browsers "fullscreen" mode


hudson-newey:
Many end users use a small screen size, and therefore cannot see a large portion of the verification grid.

We should add a prompt for users to enter full screen mode for the best experience, and provide them with a button to enter/exit full screen mode.

Additional note:

- The browser should leave full screen mode when the tab is closed and should return to full screen when the tab is re-focused

atruskie:
Probably should have a full screen button somewhere

PR #145, opened 2024-08-08
Add fullscreen toggle button

hudson-newey:
# Add fullscreen toggle button

This pull request adds an `oe-settings` component which (currently) contains a fullscreen button.

The hope is to extend this component in the future to change settings about the verification grid.

**I have not finished the tests yet, but the fullscreen component is complete, and I have outlined all the tests that I want to implement**

### Features

- Adds a fullscreen through an `oe-verification-grid-settings` component that allows the user to fullscreen the verification grid component
- Adds a message to the help dialog that we recommend using the verification grid in fullscreen mode

### Bug Fixes

- Fixes a bug where verification grid shortcut keys would not work when sub-selection was disabled (e.g. for a single item grid)

### Code Quality

- Fixes bug in the VSCode workspace which caused inconsistent formatters to run
- The `oe-data-source` now checks that the target of the `for` attribute is an `oe-verification-grid`
- Toggling the `oe-help-dialog` now listens for a `click` event (which emits a `PointerEvent` event model). This adds support for toggling the help dialog through tab-stops and improves accessibility
- Modifies the previously unused `@queryParentElement` decorator so that it can pierce shadow DOM
- Stop prettier from scanning config files & build artifacts (improving performance, and reducing bugs)

## Visual Changes

image

_A verification grid with the new fullscreen button. Note: The position is configurable and this is not the position in all verification grids_

## Related Issues

Fixes: #130 

## Additional Notes

I am using `sl-tooltip` for the fullscreen buttons tooltip.
This PR therefore adds `sl-tooltip` as a dependency

## Final Checklist

### DIFF
--- a/src/components/verification-grid/css/style.css
+++ b/src/components/verification-grid/css/style.css
@@ -35,8 +35,14 @@
   user-select: none;
   justify-content: center;
   align-items: stretch;
-  margin-top: var(--oe-spacing);
-  margin-bottom: var(--oe-spacing);
+
+  /*
+    We apply normal padding to the verification grid so that it is consistent
+    with the rest of the client, and so that the interactivity area is extended
+    slightly beyond the edge of the verification grids content
+  */
+  padding-top: var(--oe-spacing);
+  padding-bottom: var(--oe-spacing);
 
   /*
     We use a grid gap property instead of adding margin to the grid tile
--- a/src/components/verification-grid/help-dialog.ts
+++ b/src/components/verification-grid/help-dialog.ts
@@ -146,6 +146,11 @@ export class VerificationHelpDialogComponent extends AbstractComponent(LitElemen
       <p>
         Any recordings that you do not apply a decision to will be marked as "SKIP" when the results are downloaded.
       </p>
+
+      <p>
+        We highly recommend that you use the verification grids "Fullscreen Mode" by using the fullscreen icon or by
+        pressing <kbd>F11</kbd> on your keyboard.
+      </p>
     `;
   }
 
--- a/src/components/verification-grid/verification-grid.ts
+++ b/src/components/verification-grid/verification-grid.ts
@@ -294,10 +294,6 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
 
   //#region EventHandlers
   private handleKeyDown(event: KeyboardEvent): void {
-    if (!this.canSubSelect()) {
-      return;
-    }
-
     if (event.altKey) {
       // showing/hiding selection shortcuts is quite expensive because it
       // requires DOM queries
@@ -308,24 +304,6 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
       // return early here because otherwise ctrl + alt + a would select all items
       // when the expected behavior is to add item a to the sub selection
       return;
-    }
-
-    if (event.ctrlKey && event.key === "a") {
-      // we prevent default on the ctrl + A event so that chrome doesn't
-      // select all the text on the page
-      event.preventDefault();
-
-      this.subSelectAll();
-      // we return early after every keyboard shortcut because we know that
-      // we can't trigger multiple keyboard shortcuts at the same time
-      // e.g. You can't press ctrl + A and escape to select and deselect
-      // by returning early we save computing erroneous keyboard shortcuts
-      return;
-    }
-
-    if (event.ctrlKey && event.key === "d") {
-      event.preventDefault();
-      this.removeSubSelection();
     }
 
     switch (event.key) {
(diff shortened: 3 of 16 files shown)


src/components/verification-grid/css/style.css
```diff
@@ -35,8 +35,15 @@
...
-  margin-bottom: var(--oe-spacing);
+
+  /*
+    We use padding instead of margin here because the default background color
+    for fullscreen mode is black (turning margin black).
+    If margin was used, there would be black bars around the verification grid
+    appearing as if the verification grid was touching the edge of the screen
+  */
+  padding-top: var(--oe-spacing);
+  padding-bottom: var(--oe-spacing);
```
atruskie, line 46: sounds like the correct fix is to style a background color of some other element

src/components/verification-grid/help-dialog.ts
```diff
@@ -146,6 +146,11 @@ export class VerificationHelpDialogComponent extends AbstractComponent(LitElemen
       <p>
         Any recordings that you do not apply a decision to will be marked as "SKIP" when the results are downloaded.
       </p>
+
+      <p>
+        We highly recommend that you use the verification grids "Fullscreen Mode" by using the fullscreen icon or by
+        pressing <kbd>F11</kbd> on your keyboard.
+      </p>
```
atruskie, line 153: It'll do but I was thinking of a JIT prompt

src/components/verification-grid/verification-grid.ts
```diff
@@ -310,25 +310,9 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
-    }
-
-    if (event.ctrlKey && event.key === "d") {
-      event.preventDefault();
-      this.removeSubSelection();
-    }
-
     switch (event.key) {
+      // the escape button is a native browser shortcut to toggle fullscreen
+      // mode. We therefore do not want to bind to the escape key
```
atruskie, line 315: but it's still bound?

src/components/verification-grid/verification-grid.ts
```diff
@@ -346,6 +330,30 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+      case "a": {
+        if (event.ctrlKey) {
+          // we prevent default on the ctrl + A event so that chrome doesn't
+          // select all the text on the page
+          event.preventDefault();
+        }
+
+        if (canSubSelect) {
+          this.subSelectAll();
+        }
```
atruskie, line 352: when can we ever not sub-select?

atruskie (changes requested):
- I'm not convinced we need a public facing component for this.
- I'm not sure `oe-settings` is an appropriate name - is it settings for the whole app or just the verification grid?
- I'm not sure there's a need for us to be templating the component
- And all of the above naturally would lead to, why do we have a `for` attribute for this component


I didn't understand the `controls-left`/`controls-right` slots initially. Probably counts as feature creep for the scope of this issue.

src/components/verification-grid/verification-grid.ts
hudson-newey: When there is one tile in the verification grid

This was a requested change from our first round of user testing

src/components/verification-grid/css/style.css
hudson-newey: This is how I originally tried to fix this problem

Remember that we are creating styles for web components, so we can't change the `body` background color or any other element that we don't control

In my mind the `.verification-grid` is the "top most" level element that we control.

---

Additionally: The fact that the verification-grids background did not extend all the way to the edge of the screen (even in windowed mode) is a bug.

Imagine that the background color was something more distinct from the default background color (white)

You'd notice right away that the background color didn't cover the entire screen

src/components/verification-grid/verification-grid.ts
hudson-newey: The comment is incorrect

<https://stackoverflow.com/a/37759224>

I've removed the comment and left the code unchanged

src/components/verification-grid/verification-grid.ts
atruskie: Sure. Perhaps move it to the selection functions.
My gripe here is that is looked like special casing. Rather than remember to do it everywhere let's just let the selection functions take care of choosing whether or not they have an affect?

src/components/verification-grid/css/style.css
atruskie: >Remember that we are creating styles for web components, so we can't change the body background color or any other element that we don't control

> In my mind the .verification-grid is the "top most" level element that we control.

Valid, but then why are we "fixing" this at all?

You know what I think it's your comment that's misleading. It's pretty normal for an element to have some padding inside. Why talk about margins and background colors at all? That's the host's concern.

---

So, is the issue here that the actual element is being made full screen?  And because it is removed from it's host then we get a black background?

Seems like this should be the fix: https://stackoverflow.com/a/75765926/224512

src/components/verification-grid/verification-grid.ts
hudson-newey: Fixed in commit: [223bf99759deaff9c0141e82a0aa515a6e7ff2ba](https://github.com/ecoacoustics/web-components/commit/223bf99759deaff9c0141e82a0aa515a6e7ff2ba)

atruskie (approved):

Issue #137, opened 2024-08-02
Add support for a user configurable grid size option in the UI


atruskie:
Users would like to adjust the density of the grid easily while they are using it.

We already have a grid-size option, but it is set in the HTML and that is not accessible to an average user.

One user has asked to control both grid height and width but we think this adds:

- unnecessary complexity for the most common users
- and more confusion when have shrink the number of grid tiles for very small screens/phones (we can show one number being reduced to fit space easier than two, and we can flow items naturally as the screen gets smaller).

Issue #147, opened 2024-08-12
Verification grid should attempt to grow if the screen size changes


hudson-newey:
At the moment, we check to see if the verification grids size will fit in the current viewport (soon to be container with the completion of #146).

However, we do not have any functionality to grow the verification grid to the "ideal" verification grid size set by the `grid-size` attribute.

This means that the verification grid can shrink when the browser window is resized, but not grow.

We should therefore listen for the browser resize events so that we can grow the verification grid if it has previously been shrunk by the intersection observer.

Issue #146, opened 2024-08-12
Verification grid shrinking can be too eager to shrink when scrolling


hudson-newey:
At the moment, the verification grid will decrease its grid size until all the verification grid tiles can fit in the viewport.

However, this can be a problem if the user scrolls off the screen, because the intersection observer will continuiously report that the verification grid does not fit in the current viewport, and will continue to shrink it until there is only one item.

We should instead link the intersection observer to a verification grid container element that will aim to be the size of the screen (but can be limited if the host application overrides the size of the verification grid).

e.g.

```css
position: relative;

width: 100dvh;
max-width: 100%;
height: 100dvh;
max-height: 100%;
```

Therefore, if the user scrolls the verification grid out of the viewport, the intersection observer won't shrink the verification grid because it will be checking if the verification grid fits inside the container instead of checking if the verification grid size fits in the viewport.

PR #148, opened 2024-08-12
Add UI to change verification grid size

hudson-newey:
# Add UI to change verification grid size

This pull request exposes a user interface to change the verification grid size after creation

### Features

- Adds a button to change the verification grid size

### Bug Fixes

- Fix a bug where increasing the grid size after initalization would not cause the new spectrograms to generate
- Fix a bug where if the same URL was used for two grid tiles in the same position, the subject model would not update

### Code Quality

- Corrects typing of media controls `changeNumberHandlerEvent` callback to a `HTMLInputElement` `Event` instead of a `SLMenuItem` event

### Remaining Bugs / Unresolved Problems

Because this PR exposes the verification grid size to the user #146, and #147 become much more prevelant because the user can input grid sizes that (obviously) won't fit on their screen and start scrolling.

## Visual Changes

image

_A new scrollbar + label that can be used to change the verification grid size_

image

_Media controls now has a wrapper around the range inputs_

## Related Issues

Fixes: #137 

## Additional Notes

N.A.

## Final Checklist

### DIFF
--- a/src/components/verification-grid-settings/verification-grid-settings.ts
+++ b/src/components/verification-grid-settings/verification-grid-settings.ts
@@ -1,8 +1,10 @@
-import { customElement, property } from "lit/decorators.js";
+import { customElement, property, state } from "lit/decorators.js";
 import { AbstractComponent } from "../../mixins/abstractComponent";
 import { html, LitElement, unsafeCSS } from "lit";
 import { VerificationGridComponent } from "verification-grid/verification-grid";
 import { queryParentElement } from "../../helpers/decorators";
+import { ChangeEvent } from "../../helpers/advancedTypes";
+import { ifDefined } from "lit/directives/if-defined.js";
 import settingComponentStyles from "./css/style.css?inline";
 
 @customElement("oe-verification-grid-settings")
@@ -11,6 +13,10 @@ export class VerificationGridSettingsComponent extends AbstractComponent(LitElem
 
   @property({ attribute: false, type: Boolean })
   public isFullscreen = false;
+
+  /** An internal representation of the verification grids size */
+  @state()
+  private gridSize?: number;
 
   @queryParentElement({ selector: "oe-verification-grid" })
   private verificationGrid?: VerificationGridComponent;
@@ -48,7 +54,65 @@ export class VerificationGridSettingsComponent extends AbstractComponent(LitElem
     this.isFullscreen = document.fullscreenElement === this.verificationGrid;
   }
 
-  public render() {
+  private updateGridSizeState(): void {
+    if (!this.verificationGrid) {
+      throw new Error("Could not find associated verification grid component");
+    }
+
+    this.gridSize = this.verificationGrid.gridSize;
+  }
+
+  private handleGridSizeChange(event: ChangeEvent<HTMLInputElement>) {
+    if (!this.verificationGrid) {
+      throw new Error("Could not find associated verification grid component");
+    }
+
+    // Number is more strict in parsing values than parseInt
+    const inputValue = event.target.value;
+    const newGridSize = Number(inputValue);
+
+    this.verificationGrid.gridSize = newGridSize;
+
+    this.updateGridSizeState();
+  }
+
+  private gridSizeSettingsTemplate() {
+    if (!this.verificationGrid) {
+      throw new Error("Could not find associated verification grid component");
+    }
+
+    return html`
+      <sl-dropdown placement="top-start">
+        <sl-tooltip slot="trigger" content="Change the verification grid size">
+          <button
+            id="grid-size-trigger"
+            @click="${this.updateGridSizeState}"
+            class="oe-btn-secondary"
+            ?disabled="${!this.verificationGrid}"
+            aria-label="Change the verification grid size"
+          >
+            <sl-icon name="grid" class="large-icon"></sl-icon>
+          </button>
+        </sl-tooltip>
+
+        <sl-menu>
+          <label id="grid-size-label">
+            ${this.gridSize}
+            <input
+              id="grid-size-input"
+              @change="${this.handleGridSizeChange}"
+              type="range"
+              value="${ifDefined(this.gridSize)}"
+              min="1"
+              max="12"
+            />
+          </label>
+        </sl-menu>
+      </sl-dropdown>
+    `;
+  }
+
+  private fullscreenSettingsTemplate() {
     // document.fullscreenEnabled is a browser API that checks if the current
     // webpage is allowed to enter fullscreen mode
     const canEnterFullscreen = document.fullscreenEnabled;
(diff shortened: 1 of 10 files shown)


atruskie:
Observations when playing with the demo:



1. Reducing the grid size has an error where size ends up as one

image
image


2. The page scrolls around confusingly and can push the control bar off screen (separate issue I suppose)

image

3. The control is not reactive

src/components/verification-grid/verification-grid.ts
```diff
@@ -217,8 +217,20 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
 
   protected async updated(change: PropertyValueMap<this>): Promise<void> {
     const renderInvalidationKeys: (keyof this)[] = ["gridSize"];
-    const sourceInvalidationKeys: (keyof this)[] = ["getPage"];
     const tileInvalidationKeys: (keyof this)[] = ["selectionBehavior"];
+    // gridSize is a part of source invalidation because if the grid size
+    // increases, there will be verification grid tiles without any source
+    // additionally, if the grid size is decreased, we want the "currentPage"
+    // of sources to update / remove un-needed items
+    const sourceInvalidationKeys: (keyof this)[] = ["getPage", "gridSize"];
```
atruskie, line 225: I'm not sure on the scope of this... but this smells. I haven't read far enough into this diff to fully understand

I wonder if we're trying too hard here?  If extra items are added to the grid we render the page again. Any items that have the same source shouldn't change, and any items that are new get assigned a new source.

src/components/verification-grid/verification-grid.ts
```diff
@@ -217,8 +217,20 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+
+    // a full render invalidation causes all the grid tiles elements to be
+    // destroyed and re-created. This should always be a last resort
+    // we do this first so that if there is a source or tile invalidation
+    // the changes are applied to the newly created grid tile elements
```
atruskie, line 230: since when do we destroy tiles?

src/components/verification-grid/verification-grid.ts
```diff
@@ -1249,7 +1255,7 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
 
         <div class="verification-controls">
           <span class="decision-controls-left">
-            <oe-verification-grid-settings></oe-verification-grid-settings>
+            <oe-verification-grid-settings grid-size="${this.gridSize}"></oe-verification-grid-settings>
```
atruskie, line 1258: seems unnecessary - we can read the attribute from inside the element

atruskie (changes requested):
This might be related to the larger styles of the shoelace menu items, but the controls that popup should be in a panel of some kind. Especially for multiple child items, they shouldn't just hover on top without a background color.

src/components/verification-grid-settings/verification-grid-settings.ts
```diff
@@ -68,4 +128,14 @@ export class VerificationGridSettingsComponent extends AbstractComponent(LitElem
       </sl-tooltip>
     `;
   }
+
+  public render() {
+    // prettier-ignore
```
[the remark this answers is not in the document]
hudson-newey: Prettier sees this as basic string interpolation, and tries to format it as such

image

This is a (personal) code style preference, but I'd like the separate templates on different lines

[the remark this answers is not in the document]
atruskie: always document the reason for a linter suppression

src/components/verification-grid-settings/verification-grid-settings.ts
```diff
@@ -48,7 +52,63 @@ export class VerificationGridSettingsComponent extends AbstractComponent(LitElem
...
+    // the value of the input is not a number the number constructor will return
+    // NaN and we can hard fail and not change the grid size instead of
+    // changing to a grid size that the user did not expect
+    const inputValue = event.target.value;
+    const newGridSize = Number(inputValue);
+
+    // we use isFinite here to check that the value is not NaN, and that
+    // values such as Infinity are not considered as a valid grid size
+    if (!isFinite(newGridSize)) {
+      throw new Error("New grid size value could not be converted to a number");
```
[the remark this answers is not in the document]
hudson-newey: I debated this in my mind for a bit because it seems to violate the "Fail Early" principle

I agree, that this should be the responsibility of the grid component, and more specifically the attribute should check that the value provided is a valid value

src/components/verification-grid-settings/verification-grid-settings.ts
```diff
@@ -4,11 +4,15 @@ import { html, LitElement, unsafeCSS } from "lit";
...
 import settingComponentStyles from "./css/style.css?inline";
+import { ChangeEvent } from "../../helpers/advancedTypes";
 
 @customElement("oe-verification-grid-settings")
 export class VerificationGridSettingsComponent extends AbstractComponent(LitElement) {
   public static styles = unsafeCSS(settingComponentStyles);
 
+  @property({ attribute: "grid-size", type: Number })
+  public gridSize = 0;
+
```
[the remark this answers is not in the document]
hudson-newey: Replaced with `live` directive

src/components/verification-grid-settings/verification-grid-settings.ts
```diff
@@ -48,7 +52,63 @@ export class VerificationGridSettingsComponent extends AbstractComponent(LitElem
...
+          ${currentGridSize}
+          <input
+            id="grid-size-input"
+            @change="${this.handleGridSizeChange}"
+            type="range"
+            value="${currentGridSize}"
+            min="1"
+            max="12"
+          />
+        </label>
```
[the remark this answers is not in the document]
hudson-newey: This looks a lot more consistent. I like it

I will also be changing the media-controls brightness/contrast so that the look is consistent across components

src/components/verification-grid-settings/verification-grid-settings.ts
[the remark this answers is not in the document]
hudson-newey: Fixed and validated through the following tests that were added to the verification grids attribute tests

image

src/components/verification-grid/verification-grid.ts
hudson-newey: We currently destroy all the verification grid tiles during a render invalidation

I have an inline todo that we should implement a partial re-rendering + diffing algorithm instead of destroying all the tiles during a render invalidation

image

hudson-newey: Update: I implemented the proposed Diff algorithm (it was quite simple to implement). We no longer destroy grid tiles

image

src/components/verification-grid/verification-grid.ts
hudson-newey: Our source invalidation is optimized with a diff algorithm that will not re-render a tile/spectrogram if it has the same source.

By causing a source invalidation, only the new grid tiles that have been added will be affected

src/components/verification-grid/verification-grid.ts
atruskie: i am wondering why the lit `repeat` directive is not at the center of this strategy

but yes, probably a separate PR

src/components/verification-grid/verification-grid.ts
atruskie: seems like the wrong this was optimized.  A verification tile should do nothing if the same subjectwrapper is assigned to it. We shouldn't need the optimization at all. Still, out of scope. Leave a todo?

src/components/verification-grid/verification-grid.ts
hudson-newey: Related issue: #149 (linked in "TODO" comment as well)

src/components/verification-grid/verification-grid.ts
hudson-newey: In a previous PR (#104), we guaranteed that the memory address of the `SubjectWrapper` would not change throughout the program's lifetime.

So we now get this optimization through the Lit property decorators (for free).
This means we don't need the aforementioned diff optimization anymore
https://github.com/ecoacoustics/web-components/blob/main/src/components/verification-grid/verification-grid.ts#L1025-L1032

atruskie (approved):

Issue #149, opened 2024-08-13
Optimize render invalidation with efficient diff


hudson-newey:
With our current implementation, our render invalidation (`createSpectrogramElements`) causes all the grid tiles to be destroyed and re-created.

At the moment, the only change that causes a render invalidation is the change in the grid size (e.g., through attribute changes, or initialization of the grid size).

Because grid sizes are now user-configurable with #148, render invalidation can be a lot of work for very large grid sizes.

To fix this, we should implement a diff algorithm that:

- If the grid size has been decreased, hides (not destroy) the grid size delta
- If the grid size has stayed the same, have not effect
- If the grid size has increased, keep all the current grid tiles, and add more to the page (without destroying existing tiles)

_Notes for implementor: We might be able to use Lits `repeat` directive with a `subject` key, so that if the subject is the same, it doesn't mutate the tile_

Issue #115, opened 2024-08-01
Add axes, indicator, media controls, info card to verification grid (adopt common templated elements)


hudson-newey:
Some users have expressed the desire to add/remove info-cards, axes elements, etc... after the verification grid has been created.

Instead of requiring the user to add a html template element to modify how each verification grid tile looks, we should have a default template that the user can customize through the user interface.

We might consider adding this to a global settings component

atruskie:
>[Yesterday 18:59] Anthony Truskinger
>I'm starting to think we should stop using a template for the verification grid. 
 
>We might still have a template so a user can add some custom content - in other words, reduce the scope of the template.
 
>So spectrogram, axes, media controls all become non templated.
 
>But you could still template the information pane for example 
 
>[Yesterday 19:00] Anthony Truskinger
This would solve some of our event interaction issues too

PR #151, opened 2024-08-19
Move grid template into verification grid component

hudson-newey:
# Move grid template into verification grid component

This pull request provides a default template for the verification grid tiles which includes the axes, indicator, and media controls component.

### Features

- Adopts a common template for the verification grid tiles

### Bug Fixes

- Fixes a bug where the doc pages would show "No decisions found"
- Fixes leaky tests in `verification-grid.e2e.spec.ts` that were caused by some async assertions not being awaited
- Weak version matching for playwright created situations where two different versions of Playwright could be installed (resulting in CI failing to run). To fix this, Playwright versions have been locked (required pnpm-lock.json regeneration)

### Code Quality

- Virtual paging now uses Lit's optimized `repeat` directive (this was "nessecary" for this PR as it significantly improves use of signals)
- Removes render-invalidation from the verification grid component, meaning that we always use virtual paging, and will never destroy/recreate verification grid tiles (this change was nessecary so that signal lifetimes can be conseptulized easier)
- Fix missing awaits that should have been prevent on the `renderVirtualPage` method
- The verification grid's fullscreen state is now stored in a settings object signal
- Disabled the `no-complex-attribute-binding` rule in `lit-analyzer`

## Visual Changes

N.A.

## Related Issues

Fixes: #115 
Fixes: #149 (had to complete to get signals to working correctly)

## Additional Notes

This PR increases the size of the `kaleidoscope` mock dataset to allow the testing of a 20 tile grid

## Final Checklist


src/components/verification-grid/verification-grid.ts
```diff
@@ -87,6 +96,32 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
   public static readonly decisionMadeEventName = "decision-made" as const;
   private static readonly loadedEventName = "loaded" as const;
 
+  @provide({ context: verificationGridContext })
+  @property({ attribute: false })
```
atruskie, line 100: why not state

src/components/verification-grid/verification-grid.ts
```diff
@@ -87,6 +96,32 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+  public set axes(value: boolean) {
+    this.settings.axes.value = value;
+  }
+
+  /** Whether to show the indicator component in each grid tile */
+  @property({ attribute: "indicator", converter: booleanConverter })
+  public set indicator(value: boolean) {
+    this.settings.indicator.value = value;
+  }
+
```
atruskie, line 124: question: are getters still defined?

src/components/verification-grid/verification-grid.ts
```diff
@@ -87,6 +96,32 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+  @property({ attribute: "axes", converter: booleanConverter })
+  public set axes(value: boolean) {
+    this.settings.axes.value = value;
+  }
+
+  /** Whether to show the indicator component in each grid tile */
+  @property({ attribute: "indicator", converter: booleanConverter })
+  public set indicator(value: boolean) {
+    this.settings.indicator.value = value;
+  }
```
atruskie, line 123: out of scope. Is this really the solution we want anyway?

atruskie (changes requested):

src/components/verification-grid/verification-grid.ts
hudson-newey: No, which is really cool

image

src/components/verification-grid-tile/verification-grid-tile.ts
```diff
@@ -50,10 +51,16 @@ export class VerificationGridTileComponent extends AbstractComponent(LitElement)
 
   public static readonly selectedEventName = "selected" as const;
 
+  // I use type: Object here so that the lit-plugin VSCode extension will throw
+  // an error if we use the model in an attribute without dot notation
   @provide({ context: gridTileContext })
-  @property({ attribute: false })
+  @property({ attribute: false, type: Object })
```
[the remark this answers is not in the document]
hudson-newey: image

Why you don't want to use [model]="${this.model}"

https://github.com/runem/lit-analyzer/blob/master/docs/readme/rules.md#%EF%B8%8F-no-complex-attribute-binding

src/components/verification-grid/verification-grid.ts
atruskie: I'd say more confusing than not. If this is part of the public api then we probably want the getters created

src/components/verification-grid-settings/verification-grid-settings.ts
```diff
@@ -104,7 +114,7 @@ export class VerificationGridSettingsComponent extends AbstractComponent(LitElem
               type="range"
               value="${ifDefined(this.gridSize)}"
               min="1"
-              max="12"
+              max="20"
```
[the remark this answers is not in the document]
atruskie: i was thinking 36. let's not set an arbitrary limit. What if they have a huge screen?

what actually fits and fitting things to screen is a separate concern

src/components/verification-grid-tile/verification-grid-tile.ts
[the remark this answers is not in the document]
atruskie: Ok then you've probably taken the wrong lesson from that analyzer. It is saying, correctly, that binding complex object through attributes is not wise (without a converter) because attributes are all string serialized values.

If we don't use the property decorator here then we won't have an issue

src/components/verification-grid/verification-grid.ts
hudson-newey: Agreed.

This property has been removed as per a previous comment

src/components/verification-grid-settings/verification-grid-settings.ts
```diff
@@ -133,6 +143,74 @@ export class VerificationGridSettingsComponent extends AbstractComponent(LitElem
...
+                @change="${changeHandler("axes")}"
+                class="template-change-input"
+                type="checkbox"
+                ?checked="${watch(this.verificationGrid.settings.axes)}"
+              />
+              Axes
+            </label>
+          </sl-menu-item>
+        </sl-menu>
+      </sl-dropdown>
```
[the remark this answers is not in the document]
hudson-newey: Removed

atruskie (approved):

Issue #98, opened 2024-07-22
Shoelace icons are not bundled during production build


hudson-newey:
During the Vite production build, shoelace icons are not bundled as part of the `components.js` barrel file.

Be careful as this is not caught during the dev build, or docs dev server because they are served through Vite.

To properly test this functionality, you will need to move the `components.js` file out of the web components repository and test it with a fresh Vite config.

## Screenshots

![image](https://github.com/user-attachments/assets/77ab85d1-7be5-4db1-b91b-6b524eef0d6a)

PR #156, opened 2024-08-27
Fix: Icons not being included in production build

hudson-newey:
# Fix: Icons not being included in the production build

Previously, shoelace icons were not bundled during a production build, meaning that the media controls and other similar locations would have an empty placeholder where the icon should be.

This PR adds the shoelace icons to the production build by assigning each icon's SVG to a variable in the bundle's barrel file.

I have chosen this solution because

- We have to support offline usage (meaning that we can't use a CDN), and all assets must be included in the build
- I inlined the SVGs into the barrel file because shipping an additional dependency (in the form of a seperate SVG file in the `assets/` directory) increases the complexity of deploying the web components because it requires you to: correctly set up CORS for this external file, serve the external file with the correct MIME type, permissions, etc... By inlining it into the single barrel file, there is no additional overhead to using the web components

### Features

N.A.

### Bug Fixes

- Fixes shoelace icons not being included in the production build

### Code Quality

N.A.

## Visual Changes

image

_Icons remain unchanged, but I have included a screenshot in case you can spot any discrepancies_

## Documentation Examples

<https://66cd6037baf6ea88c3fb7d86--oe-web-components.netlify.app/examples/verification/single/>

## Related Issues

Fixes: #98 

## Final Checklist

### DIFF
--- a/dev/verification-grid.html
+++ b/dev/verification-grid.html
@@ -8,7 +8,7 @@
   </head>
 
   <body>
-    <oe-verification-grid id="verification-grid" grid-size="8" axes="false" indicator="false" media-controls="false">
+    <oe-verification-grid id="verification-grid">
       <template>
         <oe-info-card></oe-info-card>
       </template>
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "prismjs": "^1.29.0"
   },
   "devDependencies": {
-    "@11ty/eleventy": "3.0.0-alpha.10",
+    "@11ty/eleventy": "3.0.0-beta.1",
     "@11ty/eleventy-plugin-syntaxhighlight": "^5.0.0",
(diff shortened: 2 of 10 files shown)


atruskie (commented):
I'm not super keen on the inlining of svg icons. It would generate a lot of dom , especially for icons that are repeated in a grid. It's certainly inefficient. 

Can we do better?

https://shoelace.style/components/icon#customizing-the-default-library

hudson-newey:
Here a mock-up of an example using `registerBundledIcons`

image

Pros:
- The DOM's document no longer has a `src="<svg>xyz</svg>"` on each `sl-icon`

Cons:
- Icon imports are no longer co-located with the usage. We have to remember to update our bundled icon library every time we want to use a new icon

atruskie (approved):

Issue #150, opened 2024-08-13
Verification grid tile intersection observer should attempt to shrink/grow tiles before hiding


hudson-newey:
At the moment, the intersection observer will always hide verification grid ties that do not fit on the page.

Instead, verification grid tiles should attempt to shrink themselves if they do not fit in the verification grid until they reach the spectrograms minimum width.

Once the spectrograms minimum width is reached, it should resort to hiding the excess spectrograms that do not fit in the container/screen.

---

Additionally, if the verification grid tiles do not fill up the entire page, they should attempt to grow in size until they will not fit

PR #186, opened 2024-09-18
Feat: Grid tiles automatically size to fill grid

hudson-newey:
# Feat: Grid tiles automatically size to fill grid

A short description of the pull request

### Features

- The verification grid now uses a `display: grid` layout
- You can now specify the number of rows and columns in the verification grid through the `--rows` and `--columns` css variables (e.g. `<oe-verification-grid style="--rows: 6; --collumns: 3;" />` (this comes for free and required no extra effort)
- The verification grid no longer allows you to create a verification grid that would not fit inside the verification container
- The verification grid now **always** fills its container
- The verification controls are now **always** at the bottom of the `oe-verification-grid` element
- The verification grid now has a default size for desktop and mobile
- `oe-verification-grid-tile` components can now advertise if they are overflowing from the verification grid

### Bug Fixes

- Fix a bug where the axes svg would be larger than the spectrogram component, causing overflow
- Replaces the intersection observer with a predicates to assert that the grid tiles will fit in the grid. This fixes all issues related to the intersection observer
- When virtually measuring the axes component, it now adds padding to both sides of the label, better reflecting the real browser implementation. This fixes a bug where the axes labels could appear to be too close together (does not fix a similar font-measuring bug on Chrome)

### Code Quality

- The axes component's chrome is now overlayed on top of the wrapped element (spectrogram) instead of wrapping it. This allows the wrapped element/spectrogram to grow independently of the axes' size
- Changes the `oe-verification-grid`'s container to be a flex column instead of flex row. Better reflecting it's function of a container with elements of variable height
- Inadvertently correctly fixes the bug in playwright that would cause the spectrograms to infinately expand. This was done because the `oe-verification-grid` container is now a flex column instead of flex row
- Adds snapshot tests showing how the verification grid looks on different device types

## Visual Changes

Screenshots of the verification grid at different device sizes can be found [here](https://github.com/ecoacoustics/web-components/tree/dynamic-grid-tile-sizes/src/tests/__snapshots__)

image

## Related Issues

Fixes: #167
Fixes: #150
Fixes: #146
Fixes: #147
Fixes: #165

## Final Checklist


src/components/verification-grid/verification-grid.ts
```diff
@@ -292,6 +338,194 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+      // if we see the same one twice (we have increased, then decreased)
+      // we will decrement the grid size until we find a grid that fits
+      if (testedTargets.includes(refinedTarget)) {
+        refinedTarget -= 1;
+        continue;
+      }
+      testedTargets.push(refinedTarget);
+
+      let hasFittingCandidate = false;
+      for (let i = 1; i <= refinedTarget; i++) {
```
atruskie, line 437: ~so you're counting up from a grid-size of 1, to the target to find out what will fit?~

you counting up numbers to determine factors. Why don't you just calculate the factors as a separate step and then iterate though those?

Also we can optimize this step a bit. If we have our factor pairs we can assume the most "square" size is the most optimal... so we start of at a square root, and then work backwards. so for 12 we check 4 * 3 first, and then there exists the potential not to have to check every pair.

src/components/verification-grid/verification-grid.ts
```diff
@@ -292,6 +338,194 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+    const testedTargets: number[] = [];
+
+    // some numbers e.g. 9 have one factor such as 3x3 which creates a aspect
+    // ratio of 1.00
+    // using a 3x3 grid for a 16:9 screen is not ideal because it will create
+    // a lot of dead space
+    // therefore, we have a threshold that we have to meet. If we do not meet
+    // the threshold, we keep increasing the target until we find a grid shape
+    // that meets the similarity percentage threshold
+    const targetThreshold = 0.5;
```
atruskie, line 418: I suspect the threshold is playing havoc here. What happens if you remove this?

src/components/verification-grid/verification-grid.ts
```diff
@@ -292,6 +338,194 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+    }
+
+    const containerSize = this.gridContainer.getBoundingClientRect();
+    let refinedTarget = target;
+    const testedTargets: number[] = [];
+
+    // some numbers e.g. 9 have one factor such as 3x3 which creates a aspect
+    // ratio of 1.00
+    // using a 3x3 grid for a 16:9 screen is not ideal because it will create
+    // a lot of dead space
```
atruskie, line 414: our spectrograms stretch? do they create dead space?

src/components/verification-grid/verification-grid.ts
atruskie: Have you tried selecting factors with the most similar aspect ratio?

src/components/verification-grid/verification-grid.ts
```diff
@@ -292,6 +338,194 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+      // sometimes we can get stuck in an infinite loop where increasing grid
+      // size will result in the grid not fitting, but decreasing the grid size
+      // will result in a grid that does not meet the aspect ratio threshold
+      // to resolve this, we keep a record of the targets that we have checked
+      // if we see the same one twice (we have increased, then decreased)
+      // we will decrement the grid size until we find a grid that fits
+      if (testedTargets.includes(refinedTarget)) {
+        refinedTarget -= 1;
+        continue;
+      }
```
atruskie, line 433: i suspect this might be the cause of the infinite loop: if you have one counter decrementing, and one incrementing in alternate loops, you get a tug of war where you never get anyway. 

Try a recursive function, with one function iterating upwards, and the other downwards.

It doesn't even have to be recursive:

- if nothing fits at target size
    - increment `displacement`
    - check `target + displacement`
    - check `target - displacement`
    - if nothing fits, go to 1

src/components/verification-grid/verification-grid.ts
```diff
@@ -292,6 +338,194 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+      // I have added the condition so that if dynamic grid size gets stuck in
+      // and infinite loop it won't completely crash the browser
+      // meaning that the verification grid will soft fail to prevent all the
+      // users work being lost
+      const maxGridSize = 1_000 as const;
+      if (refinedTarget >= maxGridSize || refinedTarget <= 0) {
+        console.error("Reached maximum/minimum grid size. Recovering via a 1x1 grid.");
+        this.targetGridSize = 1;
+        return { columns: 1, rows: 1 };
+      }
```
atruskie, line 467: basically: this needs to go. this algorithm is either sufficiently tested that this case never occurs, or it never occurs and it is useless code.

src/components/verification-grid/verification-grid.ts
```diff
@@ -292,6 +338,194 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+            return { columns: 1, rows: 1 };
+          }
+
+          refinedTarget -= 1;
+          continue;
+        }
+
+        refinedTarget += 1;
+        continue;
+      }
```
atruskie, line 496: i recognize this in unhelpful but this gives me the heeby jeebies. See above comment about recursive functions

src/components/verification-grid/verification-grid.ts
```diff
@@ -292,6 +338,194 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+
+      // to find the optimal grid shape, we find what grid shape will result in
+      // the largest grid tile size
+      // this also inadvertently finds the grid shape that will best match the
+      // containers aspect ratio
```
atruskie, line 501: note: optimizing for area is basically the same as finding the set with the aspect ratio closest to 1

which is contradiction to the second sentence

src/components/verification-grid/verification-grid.ts
```diff
@@ -292,6 +338,194 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+        const proposedArea = proposedSize.width * proposedSize.height;
+
+        if (proposedArea > optimalArea) {
+          optimalGridShape = candidate;
+        }
+      }
+
+      // because we know that the candidate shapes array is not empty, we can
+      // safely assume that the optimal grid shape is not undefined
+      // and use a type cast to GridShape
```
atruskie, line 517: type cast?

src/components/verification-grid/verification-grid.ts
```diff
@@ -292,6 +338,194 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+      // containers aspect ratio
+      let optimalGridShape: GridShape = candidateShapes[0];
+      for (const candidate of candidateShapes) {
+        const optimalSize = this.tileSize(containerSize, optimalGridShape);
+        const proposedSize = this.tileSize(containerSize, candidate);
+
+        const optimalArea = optimalSize.width * optimalSize.height;
+        const proposedArea = proposedSize.width * proposedSize.height;
+
+        if (proposedArea > optimalArea) {
```
atruskie, line 510: naming alone: how can a proposal be greater than optimal?

src/components/verification-grid/verification-grid.ts
```diff
@@ -292,6 +338,194 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+          optimalGridShape = candidate;
+        }
+      }
+
+      // because we know that the candidate shapes array is not empty, we can
+      // safely assume that the optimal grid shape is not undefined
+      // and use a type cast to GridShape
+      return optimalGridShape;
+    }
+  }
```
atruskie, line 520: i suspect optimizing for area is a ed herring

src/components/verification-grid/verification-grid.ts
```diff
@@ -1195,10 +1417,11 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
       <div class="verification-container">
         <div
           id="grid-container"
+          class="verification-grid"
+          style="--columns: ${this.gridSizeN}; --rows: ${this.gridSizeM};"
```
atruskie, line 1421: just call those props columns and rows

atruskie (changes requested):
while reviewing I checked out the code and did some interactive testing. It's hard to describe  what is going on, but on more than one occasion I managed to cause the tab to lock up (in i'm assuming what was an infinite loop of some kind).

We also need some indication to the user that their desired grid size is being ignored. I.e. if I set grid-size to 32 a but am only showed 12 items then that's confusing. One way to do this would  be to set a maximum (and maybe issue a notice) on the grid-size ui. Another way could be to show the desired size and the current limit on the slider (again with a warning about not being able to show the requested amount).

This is some really great work, but we need another pass for reliability and edge cases.

src/components/verification-grid/verification-grid.ts
```diff
@@ -198,18 +214,29 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
   private selectionHead: number | null = null;
   private doneRenderBoxInit = false;
+  private paginationFetcher?: GridPageFetcher;
   private highlight: HighlightSelection = {
     start: { x: 0, y: 0 },
     current: { x: 0, y: 0 },
     highlighting: false,
     elements: [],
   };
+  private minimumGridSize: Size = { width: 180, height: 300 };
```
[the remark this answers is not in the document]
hudson-newey: Just the tile

In hindsight, it should be computed though

src/components/axes/css/style.css
```diff
@@ -1,21 +1,33 @@
 :host {
+  --axes-padding: 2rem;
+
   position: relative;
-  display: inline-block !important;
-  width: 100%;
+
+  &::part(tick),
+  &::part(grid) {
+    color: lightblue;
```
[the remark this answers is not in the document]
hudson-newey: image

I changed the color of the axes ticks to black

and moved opacity into the CSS parts. This means that users can now customize opacity through CSS parts

src/components/verification-grid/verification-grid.ts
```diff
@@ -67,6 +70,11 @@ interface HighlightSelection {
   elements: IntersectionElement[];
 }
 
+interface BreakpointGridSize {
+  screenWidth: Pixel;
+  gridSize: number;
```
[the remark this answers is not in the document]
hudson-newey: A default target once we have hit the breakpoints screenWidth

src/components/verification-grid/verification-grid.ts
hudson-newey: By optimizing for tile area, we are indirectly optimizing for aspect ratio.

---

The optimal grid size is determined by two factors:

- Conditions that **must** be met
- and conditions that we **should** to optimize

We need pre-conditions so that grid sizes such as 3x3 are completely rejected and the we optimize to 5x2

src/components/verification-grid/verification-grid.ts
hudson-newey: This is a bad comment, a 3x3 grid does not create "dead space" (because our spectrograms stretch)

The comment should read

> Using a 3x3 grid for a 16:9 screen is not ideal because the spectrograms will have to stretch a lot to fit. Resulting in spectrograms that have been warped and potentially lose information

src/components/verification-grid/verification-grid.ts
hudson-newey: Outdated comment

src/components/verification-grid/verification-grid.ts
hudson-newey: This loop finds the most optimal candidate, meaning that we check if the `propsedArea > foundOptimalArea`

In hindsight: 

We should find the optimal once and store the `optimalArea` value instead of re-computing it every iteration

src/components/verification-grid/verification-grid.ts
```diff
@@ -251,32 +277,52 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+    // however, if the verification grid does not have a `grid-size` attribute
+    // then we look at our grid size breakpoints to determine the grid size
+    // that will fit the best on the screen
+    if (this.targetGridSize) {
+      this.autoShapeGrid();
+    } else {
+      const screenSize = window.screen;
+      const relativeBreakpoint = this.gridSizeBreakpoints.findLast(
+        (breakpoint) => screenSize.width >= breakpoint.screenWidth,
+      );
```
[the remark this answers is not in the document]
hudson-newey: image

src/components/verification-grid/verification-grid.ts
hudson-newey: image

I suspect this also negates the need for a prime check

src/components/verification-grid-settings/verification-grid-settings.ts
```diff
@@ -111,7 +111,7 @@ export class VerificationGridSettingsComponent extends SignalWatcher(AbstractCom
               type="range"
               value="${ifDefined(this.gridSize)}"
               min="1"
-              max="36"
+              max="32"
```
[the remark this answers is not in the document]
hudson-newey: Related issue: https://github.com/ecoacoustics/web-components/issues/195

src/components/verification-grid-tile/css/style.css
```diff
@@ -15,9 +27,11 @@
   box-shadow: var(--oe-backdrop-shadow) var(--oe-panel-color);
   cursor: pointer;
   margin-bottom: 0;
-  min-width: 350px;
+  min-width: 300px;
+  min-height: 300px;
```
atruskie, line 31: is this tied to the minimum height in our grid calculations? comment and cross link at a minimum

src/components/verification-grid-tile/verification-grid-tile.ts
```diff
@@ -140,6 +149,15 @@ export class VerificationGridTileComponent extends SignalWatcher(AbstractCompone
...
     this.spectrogram.addEventListener("loaded", this.loadedHandler);
+
+    this.intersectionObserver = new IntersectionObserver((entries) => this.handleIntersection(entries), {
+      root: this,
+      // a threshold of zero indicates that we should trigger the callback if
+      // any part of the observed elements overflow the component
+      threshold: 0,
+    });
+    this.intersectionObserver.observe(this.slotWrapper);
+    this.intersectionObserver.observe(this.contentsWrapper);
```
atruskie, line 168: why observing both?

src/helpers/controllers/dynamic-grid-sizes.ts
```diff
@@ -0,0 +1,243 @@
...
+    this.isOverlapping.subscribe((value: boolean) => {
+      if (debounceTimeout) {
+        clearTimeout(debounceTimeout);
+      }
+
+      debounceTimeout = window.setTimeout(() => {
+        if (value) {
+          this.handleIntersection();
+        }
+      }, 5);
```
atruskie, line 41: why are we using setTimeout here?

src/helpers/controllers/dynamic-grid-sizes.ts
```diff
@@ -0,0 +1,243 @@
...
+    });
+  }
```
atruskie, line 43: is this debounce even needed? I thought signals were meant to deduplicate input?

src/components/verification-grid/verification-grid.ts
```diff
@@ -198,20 +218,16 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
   private showingSelectionShortcuts = false;
   private selectionHead: number | null = null;
   private doneRenderBoxInit = false;
+  private isOverlapping = signal<boolean>(false);
```
atruskie, line 221: so there's one signal bound to all tiles?

so the assumption here is that all tiles have uniform content and all their interaction observers fire at the same time and report the same thing - because if they don't, the values are just clobbered based on who reports last

src/helpers/controllers/dynamic-grid-sizes.ts
atruskie: ok i have looked at other sources of debounces and I get that the last invocation is meant to fire without any more invocations pushing on the pipe.... but it's still weird this is needed given the other quirks of these signals

src/helpers/controllers/dynamic-grid-sizes.ts
```diff
@@ -0,0 +1,243 @@
...
+
+  private targetElement: VerificationGridComponent;
+  private candidateShapes: GridShape[] = [];
+  private containerSize: Size = { width: 0, height: 0 };
+  private minimumGridCellSize: Size = { width: 300, height: 300 };
```
atruskie, line 48: related to styles? cross document

atruskie (changes requested):
Close. So close.

We talked offline about how this does not handle scenarios where the aspect ratio looks terrible, but the next candidate is not triggered because everything fits and target grid size in the only thing considered.

src/components/verification-grid-tile/verification-grid-tile.ts
hudson-newey: image

src/components/verification-grid/verification-grid.ts
hudson-newey: We talked about this offline, and I have made the appropriate fix

src/helpers/controllers/dynamic-grid-sizes.ts
hudson-newey: Debounce / SetTimeout has been removed

atruskie (approved):

PR #225, opened 2024-10-15
Docs: Add warning on navigation to deployment

hudson-newey:
# Docs: Add warning on navigation to deployment

## Visual Changes

image

_Chrome/Brave navigation warning_

---

image

_Firefox navigation warning_


atruskie (approved):
Maybe better if it only does it after and verification has been completed but yeah otherwise good

hudson-newey:
Updated code with the following changes to only display the warning after a decision has been made

image

image

Issue #245, opened 2024-11-12
Clicking verification button emitts incorrect "oe_tag" property


hudson-newey:
When making a verification decision, the incorrect tag model is emitted. This means that the incorrect "oe_tag" column is emitted.

It is important to note that the correct "oe_confirmed" value is emitted

**this bug does not effect any active deployments because all active deployments are single class**

hudson-newey:
This is an object reference issue and only occurs with click events

PR #246, opened 2024-11-12
fix: Incorrect verification decision tag updates

hudson-newey:
# fix: Incorrect verification decision tag updates

This pull request moves the responsability of adding the `tags` property on decision models from the verification grid component to the `Verification` model.

This means that the verification grid no longer mutates `Decision` model properties.

Additionally, verification models now update their tags by using a `addTag` method that returns a new instance of the model with the tag property updated.
This fixes a bug where all verification decisions produced by the same decision button would share the same reference, meaning that when the `tags` property was updated on a decision model, it would update the `tags` property on all subjects.

---

It is important to note that this doesn't effect any of our current deployments because they are all currently single species.

## Changes

- Fixes a bug where an incorrect value would be emitted in the `oe_tag` column
- Fixes a bug where classifications and verifications with the same tag name would overwrite each other
- Fixes a bug where progress meter segments would not reflect the order of decision buttons
- Adds a `addTag` method to `Verification` models that returns a new instance of the verification model with the `tag` property changed
- Adds tests for subject model methods `addDecision`, `removeDecision`, `hasDecision`, and `skipUndecided`
- Removes `hasTag` methods from subject model in favor of using `classifications.has()` and `verification.tag === xyz`. This fixes bugs where clashing tag text of verification and classification tasks would sometimes lead to bugs in pagination and result downloads
- Excludes the changelog produced by release-please from prettier formatting

## Related Issues

Fixes: #245

## Final Checklist


src/components/verification-grid-tile/verification-grid-tile.ts
```diff
@@ -291,26 +294,57 @@ export class VerificationGridTileComponent extends SignalWatcher(AbstractCompone
...
+
+  private progressMeterTemplate(): TemplateResult {
+    // prettier wants to format this as a single line because it thinks it is
+    // string interpolation
+    // to improve readability, I have disabled prettier for this line so that we
+    // can put each of the templates on a separate line
+    // prettier-ignore
+    return html`
+      ${this.classificationMeterTemplate()}
+      ${when(this.requiresVerification, () => this.verificationMeterTemplate())}
```
atruskie, line 344: does this change the order of meter segments? it looks like verification always appears last? Shouldn't order be kept in the same order as the decisions are templated in?

src/models/decisions/verification.ts
```diff
@@ -13,4 +13,14 @@ export class Verification extends Decision {
   public constructor(confirmed: EnumValue<DecisionOptions>, tag?: Tag) {
     super(confirmed, tag);
   }
+
+  /** returns a new verification instance with the tag property changed */
+  public addTag(tag: Tag): Verification {
```
atruskie, line 18: nit: 
```suggestion
  public withTag(tag: Tag): Verification {
```

`with` is commonly used when dealing with records and implies a new record is created

src/models/subject.ts
```diff
@@ -173,4 +167,33 @@ export class SubjectWrapper {
       ...classificationColumns,
     };
   }
+
+  private addVerification(model: Verification): void {
+    // when a verification decision is made, it usually doesn't have a tag
```
atruskie, line 172: terminology: what do you mean it _usually_ doesn't have a tag? in what case does it have a tag?

src/models/subject.ts
```diff
@@ -173,4 +167,33 @@ export class SubjectWrapper {
...
+    // on the model. This is because the tag is usually associated with the
+    // subject, which the verification button cannot know about when it makes
+    // the verification model
+    //
+    // additionally, because the verification model passed into this function
+    // will be the same object and share the same reference with all other
+    // verification models produced by the verification button, we create a new
```
atruskie, line 179: that's not a reason

src/models/decisions/verification.ts
atruskie, line 1: discussed out of band: it's probably a good idea to make all the fields of a decision immutable

atruskie (approved):

src/models/subject.ts
hudson-newey: good catch, "usually" is not correct

It will never have the `tag`

src/models/decisions/verification.ts
hudson-newey: Resolved in

https://github.com/ecoacoustics/web-components/pull/246/commits/b71c9b29fec673c33b5bb80a514d8667c0b0b9ec

Issue #139, opened 2024-08-05
Remove "history mode" and replace with single buffer


hudson-newey:
At the moment we have two buffers for incoming subjects (in the data-source), and subjects with a decision (in the verification-grid).

This no longer makes sense, and we should instead use a single buffer with a `verificationHead` and `renderHead`.

The verification head should only ever increase. Representing where the user has applied decisions up to.

The `renderHead` can be used to indicate what items are being rendered (moving this head back in the buffer will start viewing history).

---

### Notes to implementer

~Make sure that changing grid size is handled correctly~

- ~If the `renderHead` >= `verificationHead`, then the `renderHead` should not change~
- ~If the `renderHead` < `verificationHead`, then the `renderHead` should move back by the change delta. This is so that if you are on the first page of history, and change the grid size you get more items from history (instead of seeing unverified items)~

atruskie:
I think now that we have a more concrete idea of "actions that need to be completed to advance", we don't need to do anything special about page sizes.

I.e. if page size changes and we end up rendering a partial page of verified and unverified items, it doesn't matter. All the logic for when to advance (has each item has all necessary actions applied?) remains the same..

I'm not sure about your rules, clarify for me?

> If the renderHead >= verificationHead, then the renderHead should not change

Why? Users can page forward? That's one of the feature requests

> If the renderHead < verificationHead, then the renderHead should move back by the change delta.

If I understand your intent, I think this is better expressed as:

- If `renderHead < verificationHead` then fetch a page starting at index `renderHead - pageSize`, else fetch a page be doing `renderHead + pageSize`. That way we're always fetching pages either side of the zero point.

hudson-newey:
The two statements

> If the renderHead >= verificationHead, then the renderHead should not change
> If the renderHead < verificationHead, then the renderHead should move back by the change delta. This is so that if you are on the first page of history, and change the grid size you get more items from history (instead of seeing unverified items)

Are trying to address what would happen if the user changes their screen size while viewing history.

- If the user starts viewing history, then expands their window, I think that we should resolve the delta by rendering additional items that have already been verified by **moving the render head** back by the number of added tiles (no need to re-fetch data).
- If the user is not viewing history, then we should resolve the delta by **fetching** more items from the data-source.

---

I think you summarized it succinctly as

[...]

PR #250, opened 2024-11-20
refactor: Unify view and history subject array

hudson-newey:
# refactor: Unify view and history subject array

## Changes

- Refactors the verification grids subject history, current page buffers, currentPage, gridPageFetcher, and urlSourcedFetcher subject arrays into a single unified array
- Refactor repetitive test setup tasks, such as waiting for the document's load event, networkIdle state, and selector checks, into a streamlined test helper function.
- Fixes a bug where changing the grid size would cause the subject buffer to be cleared
- Fixes audio-prefetching race condition through a `AudioCachedState` enum that has a "REQUESTED" state
- Fixes a bug where pressing "skip" while viewing history would cause the grid to page
- Adds better handling for failed audio pre-fetching
- Adds tests to assert that sub-selections and subject decisions work correctly after changing the grid size

## Related Issues

Fixes: #139

## Final Checklist

### DIFF
--- a/src/components/data-source/data-source.ts
+++ b/src/components/data-source/data-source.ts
@@ -4,10 +4,12 @@ import { customElement, property, query, state } from "lit/decorators.js";
 import { booleanConverter } from "../../helpers/attributes";
 import { downloadFile } from "../../helpers/files";
 import { AbstractComponent } from "../../mixins/abstractComponent";
-import { Subject, SubjectWrapper } from "../../models/subject";
 import { UrlSourcedFetcher } from "../../services/urlSourcedFetcher";
 import { VerificationGridComponent } from "../verification-grid/verification-grid";
 import { required } from "../../helpers/decorators";
+import { PageFetcher } from "../../services/gridPageFetcher";
+import { SubjectParser } from "../../services/subjectParser";
+import { DownloadableResult } from "../../models/subject";
 import dataSourceStyles from "./css/style.css?inline";
 
 /**
@@ -82,12 +84,6 @@ export class DataSourceComponent extends AbstractComponent(LitElement) {
       return;
     }
 
-    // TODO: remove this hack that was added to fix an issue where if the
-    // data source was changed from a local file to a callback-based source
-    // the downloaded results would return the previously used local file
-    // as the basis for the results
-    const isCallbackFromUrlSourcedFetcher = this.verificationGrid?.getPage?.brand === UrlSourcedFetcher.brand;
-
     // 1. If the data source is a URL (or file served through a URL), we know
     //    all the contents of the dataset. Therefore, we need to return the
     //    entire dataset with additional columns for decisions
@@ -96,7 +92,7 @@ export class DataSourceComponent extends AbstractComponent(LitElement) {
     //    entire dataset when downloading. Therefore, we don't need to worry
     //    about joining the decisions with the original dataset, and we can just
     //    flatten the subjects and their decisions
-    if (isCallbackFromUrlSourcedFetcher) {
+    if (this.isUrlSourced()) {
       await this.downloadUrlSourcedResults();
     } else {
       await this.downloadCallbackSourcedResults();
@@ -104,12 +100,7 @@ export class DataSourceComponent extends AbstractComponent(LitElement) {
   }
 
   private async downloadCallbackSourcedResults(): Promise<void> {
-    if (!this.verificationGrid) {
-      throw new Error("associated verification grid could not be found");
-    }
-
-    const allDecisions = this.decidedSubjects();
-    const downloadableResults = allDecisions.map((model) => model.toDownloadable());
+    const downloadableResults = this.resultRows();
     const stringifiedResults = JSON.stringify(downloadableResults);
 
     const file = new File([stringifiedResults], "verification-results.json", { type: "application/json" });
@@ -127,7 +118,8 @@ export class DataSourceComponent extends AbstractComponent(LitElement) {
       throw new Error("Data fetcher does not have a file.");
     }
 
-    const results = await this.urlSourcedResultRows();
+    const downloadableResults = this.resultRows();
+
     const fileFormat = this.urlSourcedFetcher.mediaType ?? "";
 
     const originalFilePath = this.urlSourcedFetcher.file.name;
@@ -142,11 +134,11 @@ export class DataSourceComponent extends AbstractComponent(LitElement) {
     // in the format content-type/subtype; charset=encoding
     // e.g. text/csv; charset=UTF-8
     if (fileFormat.startsWith("application/json")) {
-      formattedResults = JSON.stringify(results);
+      formattedResults = JSON.stringify(downloadableResults);
     } else if (fileFormat.startsWith("text/csv")) {
-      formattedResults = new Parser().parse(results);
+      formattedResults = new Parser().parse(downloadableResults);
     } else if (fileFormat.startsWith("text/tab-separated-values")) {
-      formattedResults = new Parser({ delimiter: "\t" }).parse(results);
+      formattedResults = new Parser({ delimiter: "\t" }).parse(downloadableResults);
     } else {
       // we should never reach this condition because we checked that the file
       // format was supported when we loaded the data source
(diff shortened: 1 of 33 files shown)


src/components/verification-grid/verification-grid.ts
```diff
@@ -866,72 +891,33 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
     }
   }
 
-  private async renderHistory(historyOffset: number) {
-    const decisionStart = Math.max(0, this.subjectHistory.length - historyOffset);
-    const decisionEnd = Math.min(this.subjectHistory.length, decisionStart + this.populatedTileCount);
-    const decisionHistory = this.subjectHistory.slice(decisionStart, decisionEnd);
+  private async pageForward(): Promise<void> {
+    const proposedViewHead = this.viewHead + this.populatedTileCount;
+    this.viewHead += Math.min(proposedViewHead, this.verificationHead);
```
atruskie, line 906: first question is why? and i think the reasoning here is so that if the grid size changed while viewing history, when you get back to verification head, it will "align" again so that there's a full set of undecided items on the page

but we don't need to do that anymore.

if we page forward and the verification head is in the middle of a page, it shouldn't matter. `N` tiles before the verification head will already be done, and `M` afterwards will need to be done, but the logic for the entire set `P` is  'do all items have all actions completed?' 

While typing this out, I thought about a counter argument: then a user might have to sub-select to make new decisions for just the `N` subset.... 

I don't know. The intent here is to make this feel like a scrollable list. I think this is inconsistent.

src/components/verification-grid/verification-grid.ts
```diff
@@ -866,72 +891,33 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
-    // therefore, we have to recreate these grid tiles before we can start
-    // rendering history (if they are not present)
-    if (!this.gridTiles || this.gridTiles.length === 0) {
-      // TODO: forcing a synchronous update is hacky, and we should definitely remove it
-      this.performUpdate();
-    }
-  }
+  private pageBackward(): void {
+    const proposedHead = this.viewHead - this.populatedTileCount;
+    this.viewHead = Math.max(0, proposedHead);
```
atruskie, line 913: we could but the bounds checking into the setter - for min and max

src/components/verification-grid/verification-grid.ts
atruskie, line 906: possible overflow of subjects array?

or are just counting on the button being disabled so that this isn't called at the end?

src/components/verification-grid/verification-grid.ts
```diff
@@ -944,20 +930,21 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
       throw new Error("No paginator found.");
     }
 
-    this.subjectHistory.push(...this.currentPage);
-    this.currentPage = await this.paginationFetcher.getItems(count);
-    this.verificationBuffer = this.currentPage;
-    await this.renderVirtualPage(this.currentPage);
+    const nextBuffer = await this.paginationFetcher.getItems(count);
```
atruskie, line 933: hmm one of the points of this change is that the fetching of pages of subjects was meant to be decoupled from paging of the grid.

So, in the background we can fetch thousands of subjects asynchronously.

The subject pager also shouldn't even accept a count of items to fetch. The ecosounds APi for example ideally fetches 25 items at a time. That has little or nothing to do with the grid size

src/components/verification-grid/verification-grid.ts
```diff
@@ -1082,18 +1069,31 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
 
   //#region Rendering
 
-  private async renderVirtualPage(nextPage: SubjectWrapper[]): Promise<void> {
+  private async renderVirtualPageIndex(index: number): Promise<void> {
+    const virtualPage = this.subjectBuffer.slice(index, index + this.populatedTileCount);
+    await this.renderVirtualPage(virtualPage);
+  }
+
+  private async renderVirtualPage(pageSubjects: SubjectWrapper[]): Promise<void> {
```
atruskie, line 1077: do we need the flexibility of these overloads?

```suggestion
  private async renderVirtualPage(): Promise<void> {
```

Where the method always just renders from viewHead to populatedTileCount.

I also don't love the slice of the virtual page here. Seems unnecessary.

src/components/verification-grid/verification-grid.ts
```diff
@@ -1082,18 +1069,31 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
...
+    // spectrograms to load
+    const hasChangedSource = !pageSubjects.every((subject) => this.currentPage.includes(subject));
+    if (hasChangedSource) {
+      this.setDecisionDisabled(true);
+    }
 
     // by setting the current page, we trigger the template to render the
     // grid tiles with the new subjects
-    this.currentPage = nextPage;
+    this.currentPage = pageSubjects;
```
atruskie, line 1096: do we need the currentPage to be kept?

atruskie (changes requested):

src/components/verification-grid/verification-grid.ts
hudson-newey: Correct, this is potential overflow of the subjects array

I can add a guard to the `viewHead` setter that will clamp it to the length of the `subjects` array

It will also emit a warning if there is an attempt to set the `viewHead` to a value larger than the subject array's length

src/components/verification-grid/verification-grid.ts
```diff
@@ -179,6 +176,9 @@ export class VerificationGridComponent extends AbstractComponent(LitElement) {
   @state()
   private currentSubSelection: SubjectWrapper[] = [];
 
+  @state()
+  private _viewHead = 0;
```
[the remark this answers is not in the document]
hudson-newey: Correct, style guide advises against using a leading underscore for private variables

src/components/data-source/data-source.ts
```diff
@@ -108,7 +108,7 @@ export class DataSourceComponent extends AbstractComponent(LitElement) {
       throw new Error("associated verification grid could not be found");
     }
 
-    const allDecisions = this.decidedSubjects();
+    const allDecisions = this.verificationGrid.subjectBuffer;
```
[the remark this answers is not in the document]
hudson-newey: Discussed offline:

- For dynamic / callback paginated results, download everything the user has seen up to
- For local files / remote file data sources, return every subject back, even if they haven't seen it yet

src/components/verification-grid/verification-grid.ts
hudson-newey: Discussed offline

getPageFetcher currently has an internal array of subjects

Since we now have a combined subjects array, we should store pre-fetched subjects in this array as well

---

Changes

- getPageFetcher will still exist, but it will be a class with async methods that works over the verification grid's subject buffer
    - `populateSubjects` will populate the verification grids subjects array until we reach the pre-fetch limit
    - `cacheAudio` will async iterate over the the verification grids subjects array checking if the audio has been cached. If not, it will do the appropriate GET or HEAD request for the audio
        - cacheAudio will be automatically run if populateSubjects results in more items being appended to the subjects buffer

src/services/gridPageFetcher.ts
```diff
@@ -68,22 +77,28 @@ export class GridPageFetcher {
...
   }
 
   // during client caching, we do a GET request to the server for the
   // audio file. Therefore, requests that have already been client cached
   // have also already been server cached. We therefore, remove these
   // requests from the calculations
   private async audioCacheClient(): Promise<void> {
-    const models = this.subjectQueueBuffer.slice(0, this.clientCacheLength).filter((model) => !model.clientCached);
+    const models = this.subjectQueue
+      .slice(0, this.clientCacheLength)
```
atruskie, line 94: why always slicing from 0? shouldn't be viewHead through to viewHead + clientCacheLength?

src/services/gridPageFetcher.ts
```diff
@@ -92,11 +107,15 @@ export class GridPageFetcher {
   // audio file using ffmpeg when a file is requested.
   // by doing a HEAD request, we can warm the ffmpeg split file on the server
   private async audioCacheServer(): Promise<void> {
-    const models = this.subjectQueueBuffer.slice(0, this.serverCacheLength).filter((model) => !model.serverCached);
+    const models = this.subjectQueue
+      .slice(0, this.serverCacheLength)
```
atruskie, line 111: ditto

atruskie (approved):

Issue #108, opened 2024-08-01
Help dialog can incorrectly appear as a full page


hudson-newey:
On small screens, the help dialog can appear to be the full page size, and many users think that navigating to the verification grid is a different page.

Therefore we should:

- Ensure that there is enough padding on the top & sides of the dialog to clearly show that it is a dialog, and not a different page

Additional bugs:

- Clicking on the help dialogs scroll bar can cause the help dialog to close

atruskie:
This should be much easier when #107 is done because we have less content cram in.

Also, we need make sure size of the box never exceeds, says 80% vh/vw.

Issue #107, opened 2024-08-01
Help dialog should use animations to show how to use the verification grid


hudson-newey:
The help dialog isn't very useful in its current state as it requires the user to understand the layout of the interface, and the task that they are supposed to be doing.

We have previously tried to solve this problem by adding some information paragraphs at the top of the help dialog.
However, during user testing, we have found that most users (understandably) skip over this large clump of text.

---

We should therefore improve the verification grids help dialog by showing SVG animations on how to use the verification grid.

The animations should be different for verification and classification tasks.

The animations should be simple, low fidelity renders of the verification grid.

atruskie:
We should use a common bootstrapping paradigm: i.e. slides with a dots at the bottom

Break up the current information into a couple of slides:

1. Your task is to [verify | classify] segments - with a diagram
2. No selection guide
3. Selecting with mouse (rectangle drag, shift click, ctrl clicks)
4. keyboard short cuts
5. Configuration options and how you can find this window again

PR #265, opened 2024-12-15
[DRAFT] Animation bootstrap

hudson-newey:
# Animation bootstrap

This PR is a draft because functionality is (pretty much) complete, but code quality is not up to standard, and I have to learn more about css animation best-practices.

### Features

- Improves the help-dialog with animations

## Related Issues

Fixes: #107
Fixes: #108

## Final Checklist


src/components/bootstrap-modal/bootstrap-modal.ts
```diff
@@ -0,0 +1,262 @@
...
+      >
+        <h2 class="slide-title">${slide.title}</h2>
+        ${slide.slideTemplate}
+      </div>
+    `;
+  }
+
+  private slideFooterTemplate(): HTMLTemplateResult {
+    // TODO: Find a better way to do this
+    if (this.slides.length === 1) {
```
atruskie, line 195: yes you should - maybe set a flag when `showAdvancedDialog` is called

src/components/bootstrap-modal/slides/advanced-shortcuts/advanced-shortcuts.ts
```diff
@@ -0,0 +1,39 @@
+import { html } from "lit";
+import { KeyboardShortcut, keyboardShortcutTemplate } from "../../../../templates/keyboardShortcut";
+import { BootstrapSlide } from "../bootstrapSlide";
+
+export function advancedShortcutsSlide(): BootstrapSlide {
+  const title = "Keyboard shortcuts";
+
+  const keyboardShortcuts = [
+    { keys: ["Space"], description: "Play selected spectrograms" },
+    { keys: ["Alt"], description: "Keyboard selection" },
```
atruskie, line 10: Alt + a number key

src/helpers/directives.ts
```diff
@@ -24,3 +57,11 @@ export function* loop<T>(
     }
   }
 }
+
+/**
+ * @description
+ * Repeats a template a certain number of times
+ */
+export function repeatCount(repetitions: number, template: TimesReturn): TemplateResult {
```
atruskie, line 65: nit naming: `Count` seems redundant, suggest just `repeat`

src/helpers/themes/globalStyles.css
```diff
@@ -30,6 +30,35 @@ sl-icon {
...
+  }
+}
+
+/*
+  Swiping actions work on shoelace carousel elements, so the navigation buttons
+  on the side of carousel does not provide little functionality, but they take
+  up a lot of horizontal space (which is prime real-estate on mobile devices).
+  Therefore, I disable the navigation buttons on small width screens such as
+  mobile devices to conserve space, when alternative, more intuitive and compact
+  navigational actions can be used on these smaller devices.
```
atruskie, line 50: smart

src/helpers/themes/globalStyles.css
```diff
@@ -100,14 +129,46 @@ hr {
...
+  width: max-content;
+
+  /*
+    We have a minimum width and height so that all dialogs are clearly not a
+    part of the main content.
+    By setting a minimum width and height of 50%, we ensure that the dialog will
+    always look raised against the main content.
+  */
+  min-width: 60%;
+  min-height: 60%;
```
atruskie, line 143: suggest bumping max-width up to 80%

src/helpers/types/advancedTypes.ts
```diff
@@ -32,3 +32,5 @@ export type ChangeEvent<T extends HTMLElement> = Event & {
 
 /** A type that can be used to represent a stringified css variable */
 export type CssVariable = `--${string}`;
+
+export type Tuple<T, Length> = Array<T> & { length: Length };
```
atruskie, line 36: this is actually a pretty bad definition for a tuple - maybe a better name is FixedLengthArray ? or just not using a helper type at all

src/static/xc_whipbird.png
atruskie, line 1: This file is 125 kB.  Can we apply some post processing to shrink it?

Aim for large areas of similar colors

atruskie: Here is my attempt: 
whipbird

src/templates/keyboardShortcut.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * @description
+ * A standardized template to display keyboard shortcuts and shortcut
+ * combinations
+ */
+export function keyboardShortcutTemplate(shortcut: KeyboardShortcut): HTMLTemplateResult {
+  const isAscii = shortcut.keys.every((key) => key.charCodeAt(0) < 128);
+  const hasUpperCase = shortcut.keys.some((key: string) => key === key.toUpperCase());
+  const shouldShowShift = hasUpperCase && isAscii;
```
atruskie, line 19: perhaps we can use a shift symbol (`shift = Symbol("shift")`) in our key combinations so we don't have to rely on this heuristic - which is easy to make a mistake with. 

Or we could look for `⇧` in the key combos - they're a dev interface afterall

atruskie (changes requested):
- first slide needs a sub prompt: "Decide if the tag matches the subject"
- the negative case tag should be something other than "Incorrect-Tag". Something obviously wrong like "Car" or "cow"
- The second slide prompt needs to change. Maybe: "You can select subjects to make decisions"
- The second slide needs fewer steps. We've talked about options before, but maybe start off with the first negative tile already decided (that way there is not two repetitions of choosing false)
- Third slide, simplify prompt: "You can navigate through pages of subjects"
- Fourth slide: bug in keyboard shortcuts
image
- on narrow screens we should occupy more width the modal. See above image for keyboard shortcuts wrapping unnecessarily 
- more compact text for fourth slide: maybe "Decide true" and "Decide false" instead of "Apply a X verification"

The way we have orchestrated animations is essentially unmaintainable. It's not your fault - we've talked about the problems and possible solutions without much progress.  And we've decided to move on and not get stuck here... but ouch.

Overall great work. Just tweaks needed

src/helpers/directives.ts
hudson-newey: I agree. However, `repeat` is a built-in Lit directive

<https://lit.dev/docs/templates/directives/#repeat>

The Lit `repeat` directive is more or less an optimized `map` function (so doesn't satisfy this directives use case).

I'd like to not have conflicting directive names (because it makes importing them confusing).

I'm VERY open to alternative naming suggestions

src/templates/keyboardShortcut.ts
hudson-newey: I think I like the idea of using the ⇧ character

Maybe some sort of constant so that if we want to change the appearance of the shift character we can just change the constant

e.g.

```ts
const shiftCharacter = "⇧" as const;
```

hudson-newey:
> [atruskie](https://github.com/atruskie) (Anthony Truskinger) (3 days ago)
> The second slide prompt needs to change. Maybe: "You can select subjects to make decisions"

I don't like the wording of the suggestion because selection does not make a decision.

How about "Decisions will apply to all selected subjects"

atruskie:
> > [atruskie](https://github.com/atruskie) (Anthony Truskinger) (3 days ago)
> > The second slide prompt needs to change. Maybe: "You can select subjects to make decisions"
> 
> I don't like the wording of the suggestion because: selection does not make a decision.
> 
> How about "Decisions will apply to all selected subjects"

Great suggestion

src/helpers/directives.ts
atruskie: rails uses `5.times` a lot... but just using `times(5, ...)` would not read well

Did you see this from lit: https://lit.dev/docs/templates/directives/#range?

hudson-newey: I didn't see `range`, I think that's exactly what I was looking for

src/templates/keyboardShortcut.ts
```diff
@@ -8,18 +8,15 @@ export interface KeyboardShortcut {
...
  * A standardized template to display keyboard shortcuts and shortcut
  * combinations
  */
 export function keyboardShortcutTemplate(shortcut: KeyboardShortcut): HTMLTemplateResult {
-  const isAscii = shortcut.keys.every((key) => key.charCodeAt(0) < 128);
-  const hasUpperCase = shortcut.keys.some((key: string) => key === key.toUpperCase());
-  const shouldShowShift = hasUpperCase && isAscii;
-
   return html`
-    ${when(shouldShowShift, () => html`<kbd class="shortcut-legend">Shift</kbd> +`)}
```
atruskie: i think you misunderstood what i was saying. 

talk to with about this in the morning - i don't understand it

atruskie (approved):

Issue #229, opened 2024-10-29
Spectrogram playback can start before audio is fetched


hudson-newey:
Because our `high-frequency-time-processor` can interpolate time even if the audio element doesn't report a playback time, it is possible for the indicator to move while audio is not playing.

We should have two checks before the `high-accuracy-time-processor` emits an interpolated time.

1. Is the audio element reporting that the audio is playing?
2. Is the audio buffering
3. Has the high accuracy time deviated more than 0.5 seconds (or similar) from the reported time

This problem has raised because the spectrogram component has its own "paused" state.
We should also consider using the audio element as the source of truth for if the spectrogram is playing (perhaps pass through the audio elements properties as readonly through the spectrogram component)

---

The spectrogram playback and indicator should not move if the audio fails to fetch

hudson-newey:
Related issue:
The high-accuracy time processor can continue moving when the `<audio>` element is paused

This can cause bugs where the indicator will continue moving past the spectrogram element if the `pause` event is never received

hudson-newey:
Related bug: If you hold down the space bar, the `currentTime` will become desynced with the audio, meaning that you can make the indicator line go off the screen.

PR #317, opened 2025-03-26
fix: De-sync of audio and spectrogram currentTime

hudson-newey:
# fix: De-sync of audio and spectrogram currentTime

This commit adds rubber banding for interpolated time values and
now requires an audio element to emit a time update before time
interpolation begins.

## Changes

- If audio fails to be fetched, the spectrogram components playback will no longer break for all future audio (this caused a lot of problems in the verification grid where if one audio file failed to load, all subsequent audio playback from that spectrogram component would break.
- Fix a bug where the indicator line could go past the end of the audio
- Fix a bug where the `currentTime` (e.g. indicator line) could become desynced with the real audio time, causing the indicator line to be in the wrong position. This was easy to see if you held down the space bar on a spectrogram (try it on the old and new spectrogram creator).
- Fix a bug where certain cases would cause the Firefox `audio` elements `currentTime` not to back to $$0$$ on playback completion

## Related Issues

Fixes: #229

## Final Checklist

### DIFF
--- a/src/components/spectrogram/spectrogram.ts
+++ b/src/components/spectrogram/spectrogram.ts
@@ -621,6 +621,8 @@ export class SpectrogramComponent extends SignalWatcher(ChromeHost(LitElement)) 
       if (desync > desyncLimit) {
         console.log("Media element exceeded desync threshold (this might cause rubber banding)");
 
+        startTime -= desync;
+
         this._currentTime.value = mediaElementTime;
         this.nextRequestId = requestAnimationFrame(() => this.pollUpdateHighAccuracyTime(startTime));
 
--- a/src/components/spectrogram/spectrogram.ts
+++ b/src/components/spectrogram/spectrogram.ts
@@ -1,6 +1,6 @@
 import { LitElement, PropertyValues, html, unsafeCSS } from "lit";
 import { customElement, property, query, queryAssignedElements } from "lit/decorators.js";
-import { computed, ReadonlySignal, signal, Signal, SignalWatcher } from "@lit-labs/preact-signals";
+import { computed, ReadonlySignal, signal, SignalWatcher } from "@lit-labs/preact-signals";
 import { RenderCanvasSize, RenderWindow, Size } from "../../models/rendering";
 import { AudioModel } from "../../models/recordings";
 import { Seconds, UnitConverter } from "../../models/unitConverters";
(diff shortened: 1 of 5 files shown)


src/components/spectrogram/spectrogram.ts
atruskie, line 1: i wonder if we should safe guard the setter of _currentTime so it can never exceed proper bounds

src/components/spectrogram/spectrogram.ts
```diff
@@ -566,33 +591,59 @@ export class SpectrogramComponent extends SignalWatcher(ChromeHost(LitElement))
...
     // audio once without polling updates
-    this.currentTime.value = this.mediaElement.currentTime;
+    this._currentTime.value = this.mediaElement.currentTime;
   }
 
-  // TODO: move somewhere else
-  private nextRequestId: number | null = null;
-
   private pollUpdateHighAccuracyTime(startTime: number): void {
     if (!this.paused) {
```
atruskie, line 614: consider setting _currentTime = this.mediaElement.currentTime when not playing (the else case of this branch)

src/components/spectrogram/spectrogram.ts
```diff
@@ -160,6 +170,9 @@ export class SpectrogramComponent extends SignalWatcher(ChromeHost(LitElement))
   private readonly highAccuracyTimeBuffer = new SharedArrayBuffer(Float32Array.BYTES_PER_ELEMENT);
   private readonly currentTimeBuffer = new Float32Array(this.highAccuracyTimeBuffer);
 
+  // TODO: move somewhere else
+  private timeInterpolateFrameId: number | null = null;
```
atruskie, line 174: better name? this is a RAF reference yeah?

atruskie (approved):

src/components/spectrogram/spectrogram.ts
hudson-newey: image

src/components/spectrogram/spectrogram.ts
hudson-newey: Do you have any suggestions?

- `timeFrameRequestId`?
- `timeInterpolateFrame`? (without `Id`)
-

src/components/spectrogram/spectrogram.ts
hudson-newey: image

src/components/spectrogram/spectrogram.ts
hudson-newey: We settled on `interpolationCancelReference`

src/components/spectrogram/spectrogram.ts
```diff
@@ -566,33 +591,59 @@ export class SpectrogramComponent extends SignalWatcher(ChromeHost(LitElement))
...
+      // playing because we do not need to calculate the time elapsed to short
+      // circuit the time update if the audio element is not playing
       const timeElapsed = bufferTime - startTime;
 
-      this.currentTime.value = timeElapsed;
+      const desyncLimit = 0.1 satisfies Seconds;
+      const desync = Math.abs(mediaElementTime - timeElapsed);
+
+      if (desync > desyncLimit) {
+        console.log("Media element exceeded desync threshold (this might cause rubber banding)");
```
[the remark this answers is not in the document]
hudson-newey: 100% agree, if the user is already lagging we should not lag them more.

Just as a note though, this condition rarely hits.
It only really hits if:

- The audio element fails to fetch/play audio
- The codec is not supported in the browser
- The user holds down the space bar
- The user has a REALLY slow internet or computer which causes the byte range requests to complete slower than playback

PR #320, opened 2025-04-02
docs: Add new verification workflow

Slade-AA:
Added new verification workflow for A2O data and custom tags

### DIFF
--- a/docs-src/deployments/verification_frogs.md
+++ b/docs-src/deployments/verification_frogs.md
@@ -1,0 +1,81 @@
+---
+layout: layouts/deployment.11ty.js
+title: Frog Verification
+---
+
+<oe-verification-grid id="verification-grid" grid-size="1">
+    <oe-verification verified="true" shortcut="Q"></oe-verification>
+    <oe-verification verified="false" shortcut="W"></oe-verification>
+    <oe-verification verified="true-multiple" shortcut="E"></oe-verification>
+    <oe-verification verified="false-multiple" shortcut="R"></oe-verification>
+    <oe-verification verified="unsure" shortcut="T"></oe-verification>
+    <oe-data-source id="data-source" slot="data-source" for="verification-grid" local></oe-data-source>
+</oe-verification-grid>
+
+<script>
+const helpMessage = `
+Please input your A2O 'Authentication Token'.
+You can find your authentication token at the bottom left corner of data.acousticobservatory.org/my_account
+
+1. Go to data.acousticobservatory.org
+2. Click on "Log In" in the menu bar
+3. Log into your account using your email and password
+4. Click on your username in the top-right of the menu bar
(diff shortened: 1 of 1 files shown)


atruskie:
@hudson-newey I don't think what @Slade-AA has done will work. Maybe model this has a verification + classification mixed task?

docs-src/deployments/verification_frogs.md
```diff
@@ -0,0 +1,81 @@
+---
+layout: layouts/deployment.11ty.js
+title: Frog Verification
+---
+
+<oe-verification-grid id="verification-grid" grid-size="1">
+    <oe-verification verified="true" shortcut="Q"></oe-verification>
+    <oe-verification verified="false" shortcut="W"></oe-verification>
+    <oe-verification verified="true-multiple" shortcut="E"></oe-verification>
+    <oe-verification verified="false-multiple" shortcut="R"></oe-verification>
```
hudson-newey, line 10: `verified` accepts a limited number of values defined here:

https://github.com/ecoacoustics/web-components/blob/main/src/models/decisions/decision.ts#L5-L10

Possible values:

- `false`
- `true`
- `unsure`
- `skip`

We did this so that the `oe-verified` column has predictable values that can be ingested by anything that conforms to the annotation format.

What you probably want is `additional-tags="multiple"`

If you use `additional-tags="multiple"`, you'll have the `oe-verified` column in the downloaded results, but also have an additional `oe-multiple` column which will either be `true` or empty.

hudson-newey:
```suggestion
    <oe-verification verified="true" additional-tags="multiple" shortcut="E"></oe-verification>
    <oe-verification verified="false" additional-tags="multiple" shortcut="R"></oe-verification>
```

atruskie: If you have the permissions make the change for @Slade-AA

hudson-newey: For context, this is what the additional-tags solution would look like

image

atruskie: Yeah that seems to be what he wants

hudson-newey (approved):

hudson-newey:
Will be merging once tests pass

Just some notes on the format that you'll be getting back

You'll have some columns like so

image

the `oe-confirmed` column comes from the "true" and "false" buttons

The additional "oe-multiple" column will only be filled in if one of the buttons with the "multiple" word is pressed.

An `oe-multiple` value of `FALSE` means that the user pressed "False (multiple)" an `oe-multiple` value of `TRUE` means that the user pressed "True (multiple)".

To re-enforce the idea, if the user pressed "True" or "False" (notice that there is no "(multiple)"), the `oe-multiple` column will empty.

---

If you have any further questions / are unsure about anything else, please don't hesitate to email us.

---

This type of use case is one of the least used, so I welcome any feedback/criticism on the UI or the result format.

Issue #243, opened 2024-11-11
Uploading local dataset in Firefox ESR is broken (windows only)


hudson-newey:
Providing a local dataset through Windows Firefox ESR (version 128.4.0esr 64-bit) doens't work

It fails with the error message "Unsupported file format"

I suspect this is something to do with Firefox converting the file into a blob without a mime type. So that when we use the file constructor, it doesn't have a `file.type`

hudson-newey:
If we cannot determine the file type from the mime type we should use the file extension as a fallback.

PR #341, opened 2025-05-01
fix: Local .csv data sources not being recognised

hudson-newey:
# fix: Local .csv data sources not being recognised

Firefox ESR on Windows was reporting .csv file types as excel files. To fix this, we now check if an "excel" file type also has a .csv extension. It does, we parse the file as a csv file.

## Changes

- If an "excel" file type is uploaded with the ".csv" extension, we attempt to parse the file as a csv. This was done because Firefox ESR on Windows reports csv files as an "application/vnd.ms-excel" file type.

## Problems

We currently have tests to cover this, but CI only tests against greenfield versions of the browsers, meaning that this bug was not caught during testing / ci.

Adding ESR versions of Firefox and Chrome to CI would have caught this bug, but adding these browsers would double CI expenses for not much additional benefit.

## Related Issues

Fixes: #243

## Final Checklist

### DIFF
--- a/src/services/urlSourcedFetcher.ts
+++ b/src/services/urlSourcedFetcher.ts
@@ -77,12 +77,17 @@ export class UrlSourcedFetcher {
     } else if (this.mediaType.startsWith("text/csv")) {
       models = await csv({ flatKeys: true }).fromString(content);
     } else if (this.mediaType.startsWith("application/vnd.ms-excel") && this.file.name.endsWith(".csv")) {
-      // Windows can sometimes report the file type of a csv as an
-      // application/vnd.ms-excel file type.
+      // On Windows, Firefox uses the MIME Database to determine the media type
+      // of a file.
+      // However, when Excel is installed, it can overwrite the datatype for
+      // csv files to application/vnd.ms-excel.
+      //
       // Therefore, if the file type is reported as "excel" and the file
-      // extension is .csv, it is safe to assume that Windows is incorrectly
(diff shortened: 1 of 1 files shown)


src/services/urlSourcedFetcher.ts
```diff
@@ -76,6 +76,14 @@ export class UrlSourcedFetcher {
       models = JSON.parse(content);
     } else if (this.mediaType.startsWith("text/csv")) {
       models = await csv({ flatKeys: true }).fromString(content);
+    } else if (this.mediaType.startsWith("application/vnd.ms-excel") && this.file.name.endsWith(".csv")) {
+      // Windows can sometimes report the file type of a csv as an
+      // application/vnd.ms-excel file type.
+      // Therefore, if the file type is reported as "excel" and the file
+      // extension is .csv, it is safe to assume that Windows is incorrectly
```
atruskie, line 83: I wonder if this is less about correctness and more to do with the default application registered to open csv files. Actually I'd bet it is this, and FF uses the Windows mine database or something

atruskie (approved):

src/services/urlSourcedFetcher.ts
hudson-newey: I can replicate the bug on my computer and the default application I have for csv files is notepad.exe

hudson-newey: Updated comment

image

Issue #271, opened 2025-01-16
Offically add support for WebKit / MacOS / IOS


hudson-newey:
At the moment, there we do not tests WebKit on MacOS/IOS

Since a lot of Mac users use Safari or a variant, we should officially add support for WebKit by adding MacOS browser support linting, and by adding WebKit tests in CI.

After some preliminary testing, it seems that our current implantation is not functional on WebKit.

PR #344, opened 2025-05-01
feat: Add official support for MacOS, Firefox & Safari

hudson-newey:
# feat: Add official support for MacOS, Firefox & Safari

This PR adds CI tests for Firefox and Safari.

As a result, this PR also adds official support for Firefox and Safari.

## Changes

- Enables Firefox & Safari in CI (there was a bug where only Chrome would run)
- Changes default font from system default "sans-serif" to "Roboto"
- Standardizes a lot more styling off the theming
- Default theming is now injected through the "replace" method, improving compatibility with Playwright and has noticeable performance improvements
- Adds playwright linting rules to eslint
- Updating canvas size signals is now done through `requestAnimationFrame` to handle resize events. This means that if there is a canvas resize in under a frame, there is only one canvas resize `unitConverter` signal update. This was hard to debug because this bug only triggered on really fast systems, and often lead to flaky local tests (flaky tests typically did not trigger in CI because the CI runner is slower than my local dev environment).
- `ChromeProvider`'s now use `display: contents` meaning that they no longer have a layout box, and the `ChromeHost` is directly passed through. This fixes a lot of flexbox, grid, and layout bugs. (this is technically a breaking change, see below)

## Learnings

I learnt a lot during this PR, so I want to make some notes to share my learnings.

- CSSImportRule's must be defined in the document and cannot be created through the CSSOM CSSStyleSheet
- Windows build of Webkit doesn't come with `AudioContext` support (https://github.com/WebKit/webkit/blob/main/Source/cmake/OptionsWin.cmake#L100)
- `font-weight: bold` is not consistent across browsers
- `window.getComputedStyle` on Chrome has all of the applied css rules as property keys. Firefox and Safari do not. However, you can use the [`getPropertyValue`](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/getPropertyValue) method to accomplish a similar behavior (although it does include computed values)
- MacOS Chromium has support for `navigator.userAgentData`, however, some combination of MacOS + Chromium seen in the GitHub runners mis-reports the `userAgentData.platform` as "Windows"
- Sub-pixel rendering between different operating systems & browsers can cause snapshot tests to work. We should therefore be very cautious when using snapshot testing, and in the future we should really only use it for canvas rendering assertions.

## Unresolved Learnings

- Ctrl + Space on Brave Browser incorrectly starts playing audio. I suspect this is an upstream bug with Brave not canceling events when opening their command windows with keyboard shortcuts (e.g. ctrl + space)

## (technically) Breaking Changes

- `ChromeProvider`'s (e.g. `oe-axes`, `oe-indicator`, `oe-annotate`) no longer have a box model, and directly pass through the wrapped `ChromeHost` box model. This means that `width`, `height`, `padding`, `margin` etc... styles can no longer target `ChromeProvider`'s. If styles were applied to these elements, targeting the `oe-spectrogram` will have the same effect.

## Related Issues

Fixes: #271

## Final Checklist


atruskie (approved):
It'd be better if we didn't have so many sleeps in these tests, ideally were waiting on actual signals/promises to resolve. But that's just an idealized goal

.github/workflows/tests.yml
```diff
@@ -4,7 +4,7 @@ on: pull_request
 
 jobs:
   Run-Tests:
-    timeout-minutes: 45
+    timeout-minutes: 90
```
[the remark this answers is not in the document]
hudson-newey: No. This is mostly just for the retries and will hopefully be reverted.

Because a lot of the tests are failing, they are being retried and artificially extending this timeout.

.eslintrc
```diff
@@ -7,25 +7,38 @@
...
       "rules": {
-        "@typescript-eslint/no-floating-promises":"error"
+        "@typescript-eslint/no-floating-promises":"error",
+        "playwright/valid-describe-callback": "off",
+        "playwright/valid-title": "off",
+        "playwright/expect-expect": "off",
+        "playwright/no-skipped-test": "off",
+        "playwright/no-conditional-in-test": "off",
+        "playwright/no-conditional-expect": "off",
+        "playwright/no-networkidle": "off"
```
atruskie, line 41: maybe leave a comment for why this has been disabled?

package.json
```diff
@@ -41,7 +41,8 @@
     "fft-windowing-ts": "0.2.7",
     "lit": "3.2.1",
     "music-metadata": "11.0.0",
-    "prismjs": "1.29.0"
+    "prismjs": "1.29.0",
+    "standardized-audio-context": "^25.3.77"
```
atruskie, line 45: have we documented why we need this package?

src/components/data-source/data-source.fixture.ts
```diff
@@ -53,6 +53,11 @@ class DataSourceFixture {
       "oe-verification",
       "oe-verification-grid-tile",
     ]);
+
+    // By having an except statement here, playwright will continue running this
```
atruskie, line 57: so i think the await triggers the waiting behaviour right, not the expect? also, nit: spelling of expect

src/components/spectrogram/spectrogram.ts
```diff
@@ -510,8 +503,14 @@ export class SpectrogramComponent extends SignalWatcher(ChromeHost(LitElement))
       return;
     }
 
+    // We do not use a truthy assertion here because if the bufferedFrameRef
+    // value is zero, we still want to cancel the animation frame.
+    if (this.bufferedFrameRef !== null) {
+      window.cancelAnimationFrame(this.bufferedFrameRef);
+    }
+
```
atruskie, line 511: i'm not following the flow of logic here.

Is the purpose of this to cancel any current resize animations, and then restart them?

also it looks like we're only using one frame of the RAF...? why not just use set timeout?

atruskie (approved):

package.json
hudson-newey: Good catch, this is not actually needed and never used. I used this to test why Safari wasn't playing audio.

I'll remove the package shortly

.eslintrc
hudson-newey: image

src/components/data-source/data-source.fixture.ts
hudson-newey: Partially correct

I'll update the comment because it's not clear enough that this awaiting behavior is possible because playwright re-tries failed Locator assertions.

---

This `except` case is awaited because Playwright assertions over browser locators run in NodeJS, meaning that the assert needs to async communicate with the browser to get the JS property.

Playwright has this cool feature where [Locator assertions](https://playwright.dev/docs/test-assertions#auto-retrying-assertions) (and http request assertions) will keep re-trying until the assertion passes or the test times out (30 seconds for our default tests).

However, this only applies to Locator assertions. If I awaited an `except` that didn't use a browser Locator (e.g. `await expect(promise).toResolve()`), it would not retry.

---

Interestingly, if I didn't await this assertion, it would still re-try but the test would continue running, and the test case could even finish before the assertion completes!

This is one of my biggest grips with Playwright and can cause tests to become leaky, because the assertion will fail in another test, causing the assertion to make a completely unrelated test fail.

This is one of the major reasons why I added the eslint-playright-rules package in this PR because it will make linting fail if I forget to await an async assertion.

---

So in conclusion: Partially correct. The `await` makes playwright wait until the assertion passes.

But the reason why we can use this `assert` to wait until the verification grid is loaded because the assertion will retry if it fails. Most assertions do not retry, so this would not be possible with other assertions such as `toEqual`.

If I did not await this assertion, it would lead to leaky tests and not be guaranteed to perform (even one) assertion within the running test.

hudson-newey: image

Issue #369, opened 2025-06-12
Verification grid progress bar placement


atruskie:
It would be nice to add the progress bar for the verification grid to the top of the page.

Can this just be done with part styles? Probably not...

Maybe an attribute: `progress="top|bottom|disabled"`

hudson-newey:
Discussed offline:

Exposing grid areas is a leaky abstraction.

We decided that conditional template rendering + attributes is the best path forwards.

PR #374, opened 2025-06-13
feat: Grid attribute to change progress position

hudson-newey:
# feat: Grid attribute to change progress position

The verification grid now takes an optional attribute `progress-bar`
that can change the position of the progress bar to either "bottom"
(default), "top", or "hidden".

## Changes

- Adds `progress-bar-position` attribute to verification grid

## Discussion points

- Is `progress` the best name for this attribute? `progress` sounds like an override for where the decision head. If we are ok using a more verbose attribute name, we could use something like `progress-position`
- You cannot target the progress bar with css parts. Is this something we want to support? If so, I'll have to do some testing/refactoring to ensure that the progress bar is compatible with user configurable css (e.g. If we want `part` support, I need to ensure changing the `display` type, sizing, margin, padding, works correctly)

## Related Issues

Fixes: #369

## Visual Changes

image

_Verification grid with the progress bar on the top of the spectrogram_

---

image

_Mobile verification grid with the progress bar on top_

---

image

_Verification grid without a progress bar_

## Final Checklist

### DIFF
--- a/dev/use-cases/shoelace-host.html
+++ b/dev/use-cases/shoelace-host.html
@@ -43,7 +43,7 @@
       <sl-range></sl-range>
     </div>
 
-    <oe-verification-grid id="verification-grid" progress-bar="top">
+    <oe-verification-grid id="verification-grid" progress="top">
       <oe-verification verified="true" shortcut="Y"></oe-verification>
       <oe-verification verified="false" shortcut="N"></oe-verification>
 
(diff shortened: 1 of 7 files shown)


atruskie (approved):
Impl looks great.

I'd actually say `progress-bar-position` for the attribute. I don't expect this will get changed a lot. 

And the part discussion is interesting, but we've got no use case for it at the moment. so, YAGNI.

Issue #372, opened 2025-06-13
De-emphasize primary color in panel color


hudson-newey:
The panel color is currently quite saturated with the primary color.

E.g. staging likes like the following

![Image](https://github.com/user-attachments/assets/f615ef6e-3b60-4440-8cae-1fbc461754a8)

---

Note that this high saturated color is only present on newer monitors with an extended color gamut (e.g. P3). This highly saturated color is not currently an issue on srgb.

A lot of our users use srgb laptop screens, so we should design with both srgb and P3 color spaces in mind.

---

To complete this issue I'll have to learn a lot about color theory and srgb fallbacks for high color spaces, so if anyone has any experience / advice, I'm keen to hear suggestions.

Issue #185, opened 2024-09-17
Uniform decision button heights


hudson-newey:
At the moment, decision buttons can be different sizes, resulting in the skip button being significantly smaller, and decision buttons appearing uneven if there is a mix of decision buttons with and without shortcut keys.

hudson-newey:
We discussed this offline, and came up with some potential mock-ups

![image](https://github.com/user-attachments/assets/dafbda5b-c769-49bc-91d1-24bf1317cb70)

![image](https://github.com/user-attachments/assets/09935cd5-3c6d-49f5-b43d-116d4aa721aa)

atruskie:
Most recent design idea: 
![image](https://github.com/user-attachments/assets/fded479e-c255-40a6-b8f9-280951d51b1a)

PR #376, opened 2025-06-16
feat: Uniform buttons heights and styling

hudson-newey:
# Uniform buttons heights and styling

- Makes decision buttons consistent heights
- Joins classification and verification components together
- De-emphasizes the verification grid's panel colour
- Advanced shortcuts table now has correct spacing in-between shortcut keys
- Theme `color-mix` now uses `oklch` for more consistent theme colours
- Fixes some theming passthrough bugs (e.g. dialog elements do not inherit `--oe-font-color`, shoelace menu items do not use theming problems)

## Visual Changes

This PR is still a work in progress, so expect minor tweaks. However, I've provided some screenshots to show the rough idea.

image

_Classification task_

---

image

_Verification task with new "download results" button_

---

image

_Keyboard shortcuts dialog_

## Related Issues

Fixes: #185
Fixes: #372

## Final Checklist


src/components/decision/classification/classification.ts
```diff
@@ -205,7 +206,8 @@ export class ClassificationComponent extends DecisionComponent {
       >
         <span class="oe-pill decision-color-pill" style="background-color: var(${color})"></span>
 
-        <div class="button-text">${decision}</div>
+        <div class="button-text">${toTitleCase(decision)}</div>
```
atruskie, line 209: are we sure we want to modify this text? Isn't that a concern of the supplier of the decision element?

src/components/decision/verification/css/style.css
atruskie, line 1: cool

src/helpers/themes/globalStyles.css
```diff
@@ -12,6 +12,10 @@
 * {
   font-family: var(--oe-font-family);
   font-weight: var(--oe-font-weight);
+
+  font-optical-sizing: auto;
+  font-style: normal;
+  font-variation-settings: "wdth" 100;
```
atruskie, line 18: maybe choose `font-width` instead? It's a more normal and understood property

src/helpers/themes/theming.css
```diff
@@ -36,17 +48,23 @@
...
   --oe-background-color: hsl(var(--oe-theme-hue), var(--oe-theme-saturation), 100%);
   --oe-font-color: hsl(
     var(--oe-theme-hue),
     calc(var(--oe-theme-saturation) - 50%),
     calc(var(--oe-theme-lightness) - 90%)
   );
+  --oe-font-color-strong: color-mix(in oklch, var(--oe-primary-color), 80% black);
 
-  --oe-border-rounding: 0.5em;
+  --oe-border-rounding: 6px;
```
atruskie, line 67: i don't actually know what's best, but do we want border radius to not scale with font size?

i think we actually don't , like you have, but maybe we could look up some best practice

src/components/decision/classification/classification.ts
hudson-newey: This is the classification component's "decision enum" value e.g. `true`, `false`, `unsure`, `skip`.

If you want to provide a "tag name" (to be shown above the classification buttons), you can still provide an element in the default slot.

e.g.

```html
<oe-classification tag="dog">
	Cat (Yes I know it's different)
</oe-classification>
```

In my example, none of the user-supplied content will run through this `toTitleCase` helper.

atruskie: fair enough

src/helpers/themes/theming.css
hudson-newey: This has been something I've been curious about for a while.

My initial thoughts are:

I suspect we want either `px` or `rem` units.
`em` units are definitely not what we want because `em` units are relative to the current font size, meaning that if we had larger font, the border rounding would increase!

I'll do some research into best practices and let you know what we want

Issue #383, opened 2025-06-27
oe-data-source should take slottable downloads button


hudson-newey:
This would allow us to slot the ecosounds annotation download button inside of the data-source component

e.g.

```html
<oe-verification-grid>
  <oe-data-source>
    <baw-annotation-download></baw-annotation-download>
  </oe-data-source>
</oe-verification-grid>
```

PR #384, opened 2025-06-27
feat: Slottable oe-data-source downloads button

hudson-newey:
# feat: Slottable oe-data-source downloads button

## Changes

- Allows a slottable download button in the default `oe-data-source` slot
- Changes the resultRows() data-source component method to public because consumers might want to get the results of the verification grid for their own custom download button.

## Related Issues

Fixes: #383

## Final Checklist

### DIFF
--- a/dev/use-cases/ecosounds.html
+++ b/dev/use-cases/ecosounds.html
@@ -22,7 +22,9 @@
       <oe-verification verified="true" shortcut="Y"></oe-verification>
       <oe-verification verified="false" shortcut="N"></oe-verification>
 
-      <oe-data-source slot="data-source" for="verification-grid" allow-downloads="false"></oe-data-source>
+      <oe-data-source slot="data-source" for="verification-grid">
+        <button onclick="alert('Placeholder')">Slottable Downloads Button</button>
+      </oe-data-source>
     </oe-verification-grid>
 
     <footer>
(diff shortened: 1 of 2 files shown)


dev/use-cases/ecosounds.html
```diff
@@ -22,7 +22,9 @@ <h1>EcoSounds</h1>
       <oe-verification verified="true" shortcut="Y"></oe-verification>
       <oe-verification verified="false" shortcut="N"></oe-verification>
 
-      <oe-data-source slot="data-source" for="verification-grid" allow-downloads="false"></oe-data-source>
+      <oe-data-source slot="data-source" for="verification-grid">
+        <button onclick="alert('Placeholder')">Slottable Downloads Button</button>
```
atruskie, line 26: what if the slotted element is not clickable? i.e. it's just a span with some text?

It should still work from what i'm seeing, but the button with the onclick confuses me.

atruskie (changes requested):
No test?

Test should be, add in text to slot, click text, slot wrapper should detect click and trigger download?

dev/use-cases/ecosounds.html
hudson-newey: So you want to support two use cases:

Changing the style / text of the button provided by the verification grid, and allow a slottable button that handles it's own download events.
E.g. for Ecosounds, we probably want to download our own format.

---

So this runs into some issues we've had before with checking if a click event has already been handled by slotted content, or we should handle the click event inside the web components.

How do you propose that I handle a download button that uses it's own format & does not use the web components download format?

Inside the host (Ecosounds) event listener, I can `preventDefault` the `click` event, so it doesn't bubble to the web component event listener, but this would mean that all consumers of the web components would have to prevent default.

Issue #360, opened 2025-05-29
Media controls `for` should take element reference


hudson-newey:
This should reduce double renders for frameworks like Angular

PR #393, opened 2025-07-02
feat: element ref ability inmedia-controls "for"

hudson-newey:
# feat: element ref ability inmedia-controls "for"

## Changes

- Allows the media-controls `for` property to take a `SpectrogramComponent` element reference

Using an element ref should make the Workbench's search page much faster.

## Related Issues

Fixes: #360

## This PR was "TDD Vibe Coded"

This branch was created to test the new [Gemini CLI](https://cloud.google.com/gemini/docs/codeassist/gemini-cli), and see how far I could get by "TDD vibe coding". I just coined the term "TDD vibe coding". I don't know if it's a "formalized" version of "vibe coding", if my term is correct, or if anyone else has written a blog on coding like this.

Turns out, I could complete this issue in 10 minutes.

I originally tried telling Gemini to complete the task with some instructions on what I wanted it to do. It didn't complete the task and went off and started modifying unrelated code.

However, I achieved a much higher success rate if I coded the tests that it needed to pass first, and provided it with some commands to run targeted tests.

I used the following prompt with success. I had to interject twice because it used the wrong playwright command, and tried to re-order an event listener, which would introduce a new bug where the `for` attribute would become disassociated with the internal `spectrogramComponent` reference if assignment hard failed (e.g. in some SSR cases where `document`/`getRoot()` is not defined).

```text
I have written some tests in @src/components/media-controls/media-controls.spec.ts for some functionality that I haven't implemented yet. The media-controls component currently takes a "for" attribute that references an elements id that it is attached to. However, I also want it to take a SpectrogramComponent reference so that I can programatically assign the spectrogram to the media-controls. If you have a solution, run the tests and assert that it works. You should avoid modifying the tests as much as possible.
```

This prompt was my second attempt and took Gemini 36 seconds to complete.

### Limitations

- The Gemini CLI doesn't seem to have access to the language server, so there were some linting mistakes that it couldn't fix without first running the `pnpm lint` command
	- Possible fix: I probably have to configure the Gemini cli to use the `ts-ls` through mcp <https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/configuration.md#available-settings-in-settingsjson>

## Final Checklist

### DIFF
--- a/src/components/media-controls/media-controls.spec.ts
+++ b/src/components/media-controls/media-controls.spec.ts
@@ -1,49 +1,88 @@
 import { expect } from "../../tests/assertions";
+import { SpectrogramComponent } from "../spectrogram/spectrogram";
+import { MediaControlsComponent } from "./media-controls";
 import { mediaControlsFixture as test } from "./media-controls.fixture";
 
 test.describe("audio element communication", () => {
-  [false, true].forEach((useElementId: boolean) => {
-    test.describe(useElementId ? "using spectrogram id" : "using element reference", () => {
-      test.beforeEach(async ({ fixture }) => {
-        if (useElementId) {
-          await fixture.createWithId();
-        } else {
-          await fixture.createWithRef();
-        }
+  function elementCommunicationTests() {
+    test("should correctly attach a spectrogram component", async ({ fixture }) => {
+      // We use "evaluate" here because we want to compare browser object
+      // references which cannot be reliably passed across the nodejs/browser
+      // Playwright boundary.
+      const hasEqualReference = await fixture.component().evaluate((mediaControls: MediaControlsComponent) => {
+        // I have to use the bracket property access notation because
+        // spectrogramElement is a private field.
+        const internalRef = mediaControls["spectrogramElement"];
+
+        // This type cast here is ok because we are just removing the "null"
+        // typing because we know that the spectrogram exists in the fixture.
+        const spectrogramRef = document.querySelector("oe-spectrogram") as SpectrogramComponent;
+
+        return internalRef === spectrogramRef;
       });
 
-      // this test exists because if you don't correctly import or mount the media controls component
-      // it will still create the element tag, but with no shadow content or error indicating that mounting
-      // the component failed and will make all other tests fail because no content = a "hidden" state
-      // therefore, by having a mounting smoke test, we can ensure that this test will fail only if
-      // we have mounted the component incorrectly
-      test("creating a visible web component", async ({ page }) => {
-        const mediaControls = page.locator("oe-media-controls");
-        await expect(mediaControls).toBeVisible();
-      });
-
-      test("state before interaction", async ({ fixture }) => {
-        const isPlaying = await fixture.isPlayingAudio();
-        expect(isPlaying).toBe(false);
-      });
-
-      test("play functionality", async ({ fixture }) => {
-        await fixture.toggleAudio();
-        const isPlaying = await fixture.isPlayingAudio();
-        expect(isPlaying).toBe(true);
-      });
-
-      test("play pause functionality", async ({ fixture }) => {
-        // start playing audio
-        // by clicking the action button again, we should stop playing audio
-        await fixture.toggleAudio();
-        await fixture.toggleAudio();
-
-        const isPlaying = await fixture.isPlayingAudio();
-
-        expect(isPlaying).toBe(false);
-      });
+      expect(hasEqualReference).toEqual(true);
     });
+
+    // this test exists because if you don't correctly import or mount the media controls component
+    // it will still create the element tag, but with no shadow content or error indicating that mounting
+    // the component failed and will make all other tests fail because no content = a "hidden" state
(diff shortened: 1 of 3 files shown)


src/components/media-controls/media-controls.spec.ts
```diff
@@ -2,40 +2,48 @@ import { expect } from "../../tests/assertions";
...
-  test("play pause functionality", async ({ fixture }) => {
-    // start playing audio
-    // by clicking the action button again, we should stop playing audio
-    await fixture.toggleAudio();
-    await fixture.toggleAudio();
-
-    const isPlaying = await fixture.isPlayingAudio();
-
-    expect(isPlaying).toBe(false);
+  [false, true].forEach((useElementId: boolean) => {
```
atruskie, line 5: much like your boolean parameter in a method rule, this would be a case where i would split the test apart into different specs for readability, mainly because you're not actually varying a boolean state in the test.

src/components/media-controls/media-controls.spec.ts
```diff
@@ -2,40 +2,48 @@ import { expect } from "../../tests/assertions";
...
+
+      // this test exists because if you don't correctly import or mount the media controls component
+      // it will still create the element tag, but with no shadow content or error indicating that mounting
+      // the component failed and will make all other tests fail because no content = a "hidden" state
+      // therefore, by having a mounting smoke test, we can ensure that this test will fail only if
+      // we have mounted the component incorrectly
+      test("creating a visible web component", async ({ page }) => {
+        const mediaControls = page.locator("oe-media-controls");
+        await expect(mediaControls).toBeVisible();
+      });
```
atruskie, line 23: i wonder if this should be an assertion in the fixture - something that just runs every time?

src/components/media-controls/media-controls.spec.ts
```diff
@@ -2,40 +2,48 @@ import { expect } from "../../tests/assertions";
...
+        // start playing audio
+        // by clicking the action button again, we should stop playing audio
+        await fixture.toggleAudio();
+        await fixture.toggleAudio();
+
+        const isPlaying = await fixture.isPlayingAudio();
+
+        expect(isPlaying).toBe(false);
+      });
+    });
```
atruskie, line 46: I guess it would be good to have a test that the reference is actually set in either method, e.g. fixture.mediaControlsReference == fixture.spectrogramReference

src/components/media-controls/media-controls.ts
```diff
@@ -60,7 +60,7 @@ export class MediaControlsComponent extends WithShoelace(AbstractComponent(LitEl
 
   /** A DOM selector to target the spectrogram element */
   @property({ type: String })
-  public for = "";
+  public for: string | SpectrogramComponent = "";
```
atruskie, line 63: what happens to the attribute when an element reference is assigned? is it removed? is it set to an empty string? does it reflect at all? what should happen?

atruskie (approved):

src/components/media-controls/media-controls.ts
hudson-newey: The attribute isn't reflected, so it isn't changed.

You make a good point and I think that it should be reflected.

---

Note that this isn't unique to this change.

If you previously changed the component's `for` property, the attribute wouldn't change, meaning that the attribute could become disassociated with the property.

This might actually be the cause of some perf issues because when a framework like Angular would re-render, it'd use the attributes value and then the property would be updated.

hudson-newey: image

atruskie: so what does this actually look like when reflected?

to be clear I think this is a case where we can justifiably expect `for=""`

src/components/media-controls/media-controls.spec.ts
hudson-newey: image

src/components/media-controls/media-controls.spec.ts
atruskie: true, but there's other ways to refactor tests.

e.g. you could have a common tests function, inside the describe:


```
describe("for with property reference", () => {
  before(() => /*set up*/) 

  commonTests()
})

describe("for with attribute", () => {
  before(() => /*set up*/) 

  commonTests()
})
```

I'm not going to die on this hill, but this is how i'd structure it

src/components/media-controls/media-controls.ts
hudson-newey: image

After I've added the `reflect: true`, it now has the text `for="[object HTMLElement]"`.

As weird as this looks, I think that it's fine. It's correctly showing that it's using an element ref.

---

I think there might be a case for overriding the `SpectrogramComponent`'s `toString()` method so that it returns it's element id, so that if you assign an element reference, the id is reflected in the DOM document.

atruskie: Soni don't really care for this pr, but as a general principle for this repo I think we need a custom reflection serializer.

IMO the best solution is a valid query selector string

Issue #406, opened 2025-07-09
Suggestion of better copy for keyboard shortcuts dialog


atruskie:
## Is your feature request related to a problem? Please describe

> In the keyboard shortcuts that pop up, would the text "Select all tiles between two points" be more clear than 'Exclusive range selection'. Similarly change text from 'Toggle selection' to 'Add or remove a tile from selection".

## Describe the solution you'd like

- [ ] Adjust copy to avoid _exclusive_ and _additive_  and _toggle_ (jargon too technical)

cc @Rob-Clemens

Issue #368, opened 2025-06-12
Do not raise warning for missing spectrogram `color-map`


hudson-newey:
At the moment, we raise a console warning if the `color-map` attribute is missing with the message 'Could not find color scale "". Defaulting to grayscale'.

We should not throw this warning if the user has not provided a color map.

However, if the user does provide a color map with an empty value, we should still log the warning.

E.g. The following should error

```html
<oe-spectrogram color-map=""></oe-spectrogram>
```

atruskie:
Yup warn on invalid value, don't warn when attribute is empty

Issue #357, opened 2025-05-28
Invalid spectrogram `window-size`'s cause browser lock up


hudson-newey:
We have guard conditions to ensure that the window-size is a power of 2, and greater than 1. But for some reason, Chrome can (sometimes) bypass this guard.

This bug is inconsistent because sometimes the guard triggers, sometimes it does not.

--

My best running theory is:

This might be a race condition where the fft worker initializes before the fft processor. Meaning that the fft processor doesn't get to error before the fft worker creates an offscreen canvas (perhaps locking up the main thread).

---

I suspect the solution will be hooking into the `shouldUpdate` render hook to guard against invalid values (I think we do this in the verification grid component)

PR #413, opened 2025-07-10
fix: Edge case spectrogram attribute parsing

hudson-newey:
# fix: Edge case spectrogram attribute parsing

## Changes

- Use default "color-map" value
- Validate `window-size`'s are valid before applying
- Less technical bootstrap copy
- Improved wrapping for button group overflow (not finished)
- Verification finished copy is now centered instead of in the top left
- Fixes bug where buttons could get stuck in disabled state

## Related Issues

Fixes: #368 (Do not raise warning for missing spectrogram color-map)
Fixes: #357 (Invalid spectrogram window-size's cause browser lock up)
Fixes: #406 (Suggestion of better copy for keyboard shortcuts dialog)

## Final Checklist


src/components/decision/verification/css/style.css
```diff
@@ -34,6 +34,14 @@
   }
 }
 
+.decision-group {
+  /*
+    By using box-decoration-break, it will round the edges if the content
+    wraps.
+  */
+  box-decoration-break: clone;
```
atruskie, line 42: i don't think we ever want a group to split. will a group ever be smaller than screen width?

atruskie (approved):

src/components/decision/verification/css/style.css
atruskie: Valid and fair

Issue #129, opened 2024-08-02
Add support for sub-selecting using the arrow keys


hudson-newey:
Users have expressed that they want to be able to move the sub-selection head by using the arrow keys.

We should mimic the behavior of Windows' explorer to make the behavior as intuitive as possible.

This new behavior would conflict with the current behavior of the arrow keys which are used to navigate around in the history. Perhaps we should use `Page Up` and `Page Down` to navigate between pages in history.

atruskie:
The ask is to move the selection with cursor with arrow keys, technically that is changing the sub-selection.

Ctrl/shift semantics should be supported - additive selection.

hudson-newey:
Pressing right or down without a selection should put your cursor on the first tile.

Stretching this thinking further, if you press left or up, the selection cursor should be put on the last tile.

---

Note that care should be given to maintain the tab index of the selection head.

If you use the arrow keys to move the selection head, the tab index target should also change so that users can still navigate with tab + enter selection.

Issue #93, opened 2024-07-19
Refactor out selection handling in verification grid to reactive controller


hudson-newey:
The verification grid component is very large and handles a lot of functionality

To help with separation of concerns, increase code readability, and reduce the risk of bugs/regressions, we should create a reactive controller that the verification grid can consume.

This reactive controller would handle selection events, sub-selection, and the selection highlight box

PR #430, opened 2025-07-21
feat: Arrow key sub-selection

hudson-newey:
# feat: Arrow key sub-selection

## Changes

- Refactor selection logic into common method
- Bifurcate `selectionHead` into `selectionHead` & `focusHead`
- Change paging shortcuts to `pageUp` & `pageDown`
- Allow selecting with arrow keys
- Add support for shortcut modifiers with highlight selection

## Related Issues

Fixes: #129

## Final Checklist


src/components/verification-grid/verification-grid.ts
atruskie, line 1: it might be worth breaking this file up a bit. Like all the selection logic in one file, and all the keyboard and mouse handling in another file

src/components/verification-grid/verification-grid.ts
```diff
@@ -681,6 +814,12 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
     // related to: https://stackoverflow.com/a/78872316
     if (event.key === ESCAPE_KEY) {
       this.removeSubSelection();
+
+      // We reset the selection head so that if the user deselects all of the
+      // tiles (e.g. through the esc key), the next shift click will start a new
+      // range selection instead of starting from the old range selection
+      // position.
+      this.resetSelectionHead();
```
atruskie, line 818: do we need a preventDefault for escape?

src/components/verification-grid/verification-grid.ts
```diff
@@ -817,54 +960,40 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
     this.showingSelectionShortcuts = false;
   }
 
-  private handleSelection(selectionEvent: SelectionEvent): void {
+  private handleTileSelection(selectionEvent: SelectionEvent): void {
```
atruskie, line 959: which events trigger this callback? mouse? everything?

src/components/verification-grid/verification-grid.ts
```diff
@@ -305,31 +330,60 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
...
-  private selectionHandler = this.handleSelection.bind(this);
+  private selectionHandler = this.handleTileSelection.bind(this);
   private decisionHandler = this.handleDecision.bind(this);
 
   public subjects: SubjectWrapper[] = [];
 
   private _loaded = false;
   private decisionHeadIndex = 0;
   private viewHeadIndex = 0;
+  private isCurrentAutoSelected = false;
```
atruskie, line 341: needs some documentation or a different name.

Is this the field that tracks if we're in decide one at a time mode? I think it is.

src/components/verification-grid/verification-grid.ts
```diff
@@ -933,6 +1054,158 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
...
+    if (!this.canSubSelect()) {
+      return;
+    }
+
+    this.isCurrentAutoSelected = false;
+
+    let selectionBehavior = this.selectionBehavior;
+    if (selectionBehavior === "default") {
+      selectionBehavior = this.isMobileDevice() ? "tablet" : "desktop";
+    }
```
atruskie, line 1074: seems like a good candidate for a getter

src/components/verification-grid/verification-grid.ts
```diff
@@ -933,6 +1054,158 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
...
+
+    const selectionIndex = tileIndices;
```
atruskie, line 1076: inline this; name is misleading and you don't use it anywhere else

src/components/verification-grid/verification-grid.ts
```diff
@@ -933,6 +1054,158 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
...
+      this.removeSubSelection();
+    }
+
+    const iterableIndices = Array.isArray(selectionIndex) ? selectionIndex : [selectionIndex];
+    for (const tileIndex of iterableIndices) {
+      // If the "focus" selection behavior is set, we want to focus the tile but
+      // not select it.
+      if (focus) {
+        this.focusTile(tileIndex);
+        this.focusHead = tileIndex;
```
atruskie, line 1094: is there a reason focusTile doesn't set focusHead?

src/components/verification-grid/verification-grid.ts
```diff
@@ -933,6 +1054,158 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
...
+
+        this.addSubSelectionRange(this.rangeSelectionHead, selectionTail);
+      } else if (additive) {
+        this.selectTile(tileIndex);
+        this.rangeSelectionHead = tileIndex;
+      } else {
+        // if we reach this point, we know that the user is not performing a
+        // range selection because range selection performs an early return
+        this.toggleTileSelection(tileIndex);
+        this.rangeSelectionHead = tileIndex;
```
atruskie, line 1109: finding this section a little hard to read, maybe some examples in comments would help?

i think range is for a shift+click, and additive is for ctrl+click.

what happens when both range and additive are true?, e.g. ctrl+shift click - or is range simply the priority? Or does range imply additive?

src/components/verification-grid/verification-grid.ts
```diff
@@ -933,6 +1054,158 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
...
+    if (value !== null) {
+      this.processSelection(value, refinedOptions);
+    } else {
+      this.removeSubSelection();
+    }
+  }
+
+  private selectionHeadUndecided(selectionIndex: number): void {
+    let firstUndecidedTile: number | null = null;
+    const nextUndecidedTile = Array.from(this.gridTiles).findIndex((tile, index) => {
```
atruskie, line 1135: maybe a comment to the effect of: "find both the first undecided tile, and the next undecided tile after the current selection. first undecided will be different if the user has undecided tiles before the current selection"

src/components/verification-grid/verification-grid.ts
```diff
@@ -933,6 +1054,158 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
...
+      if (index < selectionIndex) {
+        if (firstUndecidedTile === null && !tile.taskCompleted) {
+          firstUndecidedTile = index;
+        }
+
+        return false;
+      }
+
+      return !tile.taskCompleted;
+    });
```
atruskie, line 1151: i this case i'd probably favour literate code over performant code:

```
fistUndecidedTile = this.getFirstUndecidedTile();
nextUndecidedTile = this.getNextUndecidedTile();
```

src/components/verification-grid/verification-grid.ts
```diff
@@ -933,6 +1054,158 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
...
+    };
+
+    if (value !== null) {
+      this.processSelection(value, refinedOptions);
+    } else {
+      this.removeSubSelection();
+    }
+  }
+
+  private selectionHeadUndecided(selectionIndex: number): void {
```
atruskie, line 1133:
```suggestion
  private moveSelectionHeadToNextUndecided(selectionIndex: number): void {
```

src/components/verification-grid/verification-grid.ts
```diff
@@ -1015,37 +1296,55 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
...
+    // This is done so that the correct focus order is maintained, and any
+    // future selection order behavior is automatically implemented.
+    const isNegativeSelection = highlightWidth < 0;
+    if (isNegativeSelection) {
+      intersectingTiles.reverse();
+    }
+
+    for (const tile of intersectingTiles) {
+      this.processSelection(tile.index, options);
+    }
```
atruskie, line 1324: process selection takes an array of indices, perhaps you input an array of maybe reversed indices instead of doing the outer loop here? Did I understand this right?

src/components/verification-grid/verification-grid.ts
```diff
@@ -1090,6 +1389,7 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
     const proposedViewHead = this.viewHead + this.populatedTileCount;
     this.viewHead = proposedViewHead;
     this.removeSubSelection();
+    this.resetSelectionHead();
```
atruskie, line 1392: these two method are called together a lot? Maybe a helper method like `clearSelection()` wherever they both occur

src/components/verification-grid/verification-grid.ts
```diff
@@ -933,6 +1054,158 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
...
+   */
+  private processSelection(
+    tileIndices: number | number[],
+    { additive = false, toggle = false, range = false, focus = false }: SelectionOptions = {},
+  ): void {
+    if (!this.canSubSelect()) {
+      return;
+    }
+
+    this.isCurrentAutoSelected = false;
```
atruskie, line 1069: why is this always false here?

Maybe because anytime this method is called, it is for processing more than one selections?

Maybe any selection event (one, many, none) resets this behaviour, and making a decision when selection size is 1, turns it back on again?

src/components/verification-grid/verification-grid.ts
atruskie, line 1: on the whole, really well done, but some of this is still hard to follow

atruskie (changes requested):
Amazing attention to detail, well done.

Mainly comments suggested. Functionality behaviours suggested in previous comments, which you've already addressed

src/components/verification-grid/verification-grid.ts
hudson-newey: I agree on that philosophy

However, you'll have to explain why this is unreadable, (as the author), this seems very flat and and simple to me.

src/components/verification-grid/verification-grid.ts
hudson-newey: Good catch, it does, so this assignment was a boop

src/components/verification-grid/verification-grid.ts
hudson-newey: Everything at the grid tile level

So alt + number selection, and click selection

Anything that involves the interaction of multiple tiles in a selection event (e.g. highlight selection or arrow selection is not handled by this

(I admit this is not the best separation of concerns)

---

The guard is duplicated in the `processSelection` helper, and is therefore not needed.

If I remove the guard, I think the event handler + add JSDoc's, it becomes a bit cleaner

image

atruskie: 💯comment helps a lot when trying to trace the behaviours

src/components/verification-grid/verification-grid.ts
hudson-newey: I like the idea, I've had this planned for a long time in a separate arch issue

https://github.com/ecoacoustics/web-components/issues/93

I'll probably get around to it sometime, but I think it's a big refactor, especially to put in this PR

atruskie: yup, leave for another time

src/components/verification-grid/verification-grid.ts
```diff
@@ -302,34 +327,67 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
...
+  }
+
+  private set rangeSelectionHead(value: number | null) {
+    this._rangeSelectionHead = value;
+
+    // When unsetting the selection head (e.g. with ESC key) we want to keep the
+    // focus head at the same index, so that if you start moving again
+    if (value !== null) {
+      this.focusHead = value;
+    }
```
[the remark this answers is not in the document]
hudson-newey: *Settings the selection head to null maintains the focus heads position

This is how window's explorer works.

E.g.

1. Click on an item to select it
2. Deselect the item by clicking on an empty space
3. Observe that the focus head is still on that same tile
	a. There is a dotted outline
	b. You can move this focus head with ctrl + arrows
	c. If you press the arrow keys, the selection starts from this focus head

I think the confusion came from an incomplete comment.

Attached below is the updated source:

image

src/components/verification-grid/verification-grid.ts
hudson-newey: image

src/components/verification-grid/verification-grid.ts
hudson-newey: image

src/components/verification-grid/verification-grid.ts
hudson-newey: It always sets to false, because this method is called as the result of human action (e.g. clicking on a tile, making a highlight, alt, tab selection).

(This is a bit hacky), but if the selection was made as a result of auto-selecting, `isCurrentAutoSelected` gets immediately set after this method is complete.

image

This is actually a purposeful design because (from my thought pattern):

The user should only be put in "single selection" mode after a single tile decision (opposed to selecting a single tile).

This is done so that cases where the user manually selects the last undecided tile on a page, the first tile is not automatically selected, because they are not in "single decision mode".

To be automatically put into "single decision mode" on the next page, the last tile on the previous page must have been auto selected.
Manually selecting one tile shouldn't cause the the next page's tile to be automatically selected.

src/components/verification-grid/verification-grid.ts
hudson-newey: Ctrl + click is "toggle" because ctrl + clicking on a currently selected tile will de-select it.

"additive" mode is used for the highlight selection box (although it probably shouldn't be, because in Window's explorer ctrl + highlight can toggle/deselect items).

src/components/verification-grid/verification-grid.ts
```diff
@@ -314,22 +347,116 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
   private decisionHeadIndex = 0;
   private viewHeadIndex = 0;
 
+  /**
+   * "single decision mode" will automatically advance the selection head if:
+   *    1. There is only one tile selected
+   *    2. All tasks on the selected tile is completed
+   *
+   * You can enter this mode at any time, by simply completing all tasks on a
+   * single tile.
```
atruskie, line 356:
```suggestion
   * A user can enter this mode at any time by selecting just one tile.
   * They remain in the mode by completing all tasks on that singly selected tile,
   * at which point the selection is advanced.
```

src/components/verification-grid/verification-grid.ts
```diff
@@ -314,22 +347,116 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
...
+  private get rangeSelectionHead() {
+    return this._rangeSelectionHead;
+  }
+
+  private set rangeSelectionHead(value: number | null) {
+    // Note that the bounds check upper limit is the populated tile count
+    // subtract one.
+    // This is because the populatedTileCount is indexed from 1 while the range
+    // selection value is indexed from 0.
+    const upperBound = this.populatedTileCount - 1;
```
atruskie, line 394: this could reuse `lastTileIndex`

src/components/verification-grid/verification-grid.ts
```diff
@@ -631,17 +785,59 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
     // the ctrl key if the user is on a Mac
     // https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey
     const isHoldingCtrl = hasCtrlLikeModifier(event);
+    const selectionOptions: SelectionOptions = {
+      toggle: isHoldingCtrl,
+      range: event.shiftKey,
+    };
+
+    const keySelectionHandler = isHoldingCtrl ? this.focusTile.bind(this) : this.processSelection.bind(this);
```
atruskie, line 793:
```suggestion
    // with control, only move the focus, don't change the selection
    const keySelectionHandler = isHoldingCtrl ? this.focusTile.bind(this) : this.processSelection.bind(this);
```

src/components/verification-grid/verification-grid.ts
```diff
@@ -314,22 +347,116 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
...
+    // action.
+    const proposedIndex = this.focusHead - this.columns;
+    if (proposedIndex < 0) {
+      return this.focusHead;
+    }
+
+    return proposedIndex;
+  }
+
+  private get focusDownIndex(): number {
```
atruskie, line 434: these are great abstractions but _focus_ is a verb, and it makes it sound like the methods are doing something.

suggestion `next[Left|Right|Up|Down]Index`

src/components/verification-grid/verification-grid.ts
```diff
@@ -672,6 +873,23 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
...
+      // However, we have to preventDefault on the keydown because in some
+      // browsers (e.g. Chrome), pressing escape while the page is loading will
+      // cause the browser to stop page load.
+      // Therefore, if the user starts selecting the grid tiles before the page
+      // load finishes, and presses escape to de-select tiles, we don't want to
+      // cancel the page load.
+      case ESCAPE_KEY: {
+        // If the user has nothing selected, we still allow canceling the page
+        // navigation by pressing escape.
+        if (this.currentSubSelection.length > 0) {
```
atruskie, line 888: this is technically correct, but I actually thing we should just be bad citizens here and always prevent default if we handle it. Can you imagine trying to track down this bug?

src/components/verification-grid/verification-grid.ts
```diff
@@ -933,6 +1141,110 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
...
+    if (this.userSelectionBehavior === "desktop" && !toggle && !additive) {
+      this.removeSubSelection();
+    }
+
+    const iterableIndices = Array.isArray(tileIndices) ? tileIndices : [tileIndices];
+    for (const tileIndex of iterableIndices) {
+      if (range) {
+        // if the user has never selected an item before, the multiSelectHead will be "null"
+        // in this case, we want to start selecting from the clicked tile
+        this.rangeSelectionHead ??= tileIndex;
```
atruskie, line 1175: i think if in this case, if existing selection is none (head is null) it should just be treated as 0, and we should select from first tile to current tile index

hudson-newey: Fixed in

[`eafd840` (#430)](https://github.com/ecoacoustics/web-components/pull/430/commits/eafd8409c38c25e7e4522b356d317ebcb1e15357)

Issue #432, opened 2025-07-24
Add tag choose task


atruskie:
## Is your feature request related to a problem? Please describe

Some users, when performing verification, want to take the opportunity to supply an alternate tag after making a false verification - if a skilled user is there, then don't get them back.

## Describe the solution you'd like

A way to prompt a user to supply an alternate tag if the verify false.

## Describe alternatives you've considered

Linking out to another interface for compound tasks.

## Additional context

User requested feedback

atruskie:
The UI for this will be

- another decision button
- when pressed a text input will be shown, focused
- a user will be able to type for completion to search for the tag source
- press enter to confirm
- chosen tag will apply to selection, like any other decision

hudson-newey:
Adjustment to original spec:

Our original spec provided a single subject to the `when` predicate.

`when="(subject) => subject.verification == 'false'"`

However, this doesn't work when multiple subjects are selected.

To fix this, I propose `when` should take an array of subjects as the first callback argument

```html
<oe-tag-prompt
	when="(subjects) => subjects.every((model) => model.verification === 'false')"
></oe-tag-prompt>
```

Issue #264, opened 2024-12-12
Selection highlight box can become misaligned with the cursor


hudson-newey:
In extreme circumstances, the selection highlight box can become misaligned with the tip of the cursor.

All edge cases are non-permanent and can be fixed by stopping the selection highlight.

Identified misalignment bugs:

- Misalignment after moving the cursor VERY fast
- Misalignment after moving the browser (using shortcut keys) during selection event
- Misalignment after resizing the browser (using shortcut keys) during selection event

Issue #191, opened 2024-09-19
Add indicator color for "Unsure" Decision


nellihol:
Stakeholder D wanted to see colour for "unsure" decisions, similar to red=false, green=true. Suggestion: yellow.

atruskie:
Yeah, this is a left over from the task meter implementation.

We don't want to use a colour for unsure or skip - we don't want to overload the colors. 

I'm thinking some other kind of indication:

- some kind of hatching or pattern (in the same color as the class)
- an icon?
- or text?

An icon is probably the cleanest looking

PR #435, opened 2025-07-28
feat: Compound tasks

hudson-newey:
# feat: Compound tasks

## Changes

- Adds `oe-tag-prompt` component
- Adds internal `oe-typeahead` component
- Adds `oe_new_tag` column to downloaded data source
- Adds `newTag` property to subject wrapper models
- Adds colors for "skip" and "unnsure" decisions (This adds support for non background-color decision indicators. Meaning host applications can now do something like `--verification-true: url("checkmark.png");`
- Selection highlighting now runs at 60FPS on high-end machines (use `translate: transform` instead of `left` + `top` positioning which causes layout recalculation)
- Keyboard hints now have "squarish" padding
- Adds `aria-keyshortcuts` attribute to decision buttons & tile shortcuts
- Changes default "skip" shortcut key to "s"
- `SubjectWrapper` can now have an explicitly `null` tag
- Decisions can now be conditionally marked as "not required" by using the decision button's `when` predicate
- `oe-verification-tile` component events are now `oe-` namespaced. This is not a breaking change because `oe-verification-tile` is not user-facing and is an internal component, so consumers should not be listening to these events
- verification grid's sub selection logic has been cleaned up so there is only one source of truth
- "red", "green" and "gray" colors have been remove from the color brewer set so that they don't conflict with "true", "false" and "disabled" semantic colors
- Decisions on `SubjectWrapper` are no longer stringly typed (they are hard-restricted to the enum)

### Breaking Changes

- `decision-made` verification event now emits a map of `SubjectWrapper` -> `{ change: { decisions: Decision[] } }` models. This replaces emitting just `SubjectWrapper`'s, and is needed for compound tasks since it's hard to efficiently determine if the `SubjectWrapper` emitted in the `decision-made` event was the result of a `verification` update or a `newTag` change.

### Bug Fixes

- Fixes background bleed in tile progress meter
- Fixes bug where moving the cursor really fast during selection highlight would cause mis-alignment
- Fixes `callbackConverter` `AttributeConverter` not handling the `null` case

### Not Implemented

- Ability to pass in `<option>` elements into the typeahead / `tag-prompt` components

## Discussion Points

- We talked about prepending recently used tags to the start of the `tag-prompt` component. However, I think this should be the responsibility of the host application because it's hard for our web components to know when the recently used tags should be invalidated.
	- If the user is on a shared computer, and logs out of Ecosounds, we probably want to invalidate their recently used tags. But this would not be possible on in our web components without creating a lot of new api's/methods
	- Perhaps we want to prepend the tag prompt with the most used tags in a project (perhaps by doing an initial filter query on Ecosounds)

## Visual Changes

image

_Verification grid where first tile has completed its "Tag Correction" task_

---

image

_"Tag Correction" popover + typeahead_

---

image

_New classification decision colors_

## Related Issues

Fixes: #432 (Add tag choose task)
Fixes: #439 (Add indicator color for "skip" decisions)
Fixes: #191 (Add indicator color for "Unsure" Decision)
Fixes: #264 (Selection highlight box can become misaligned with the cursor)

## Final Checklist


src/components/verification-grid/verification-grid.ts
```diff
@@ -322,6 +324,14 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
     return availableTiles - visibleSubjectCount;
   }
 
+  private get behavingSubSelection(): SubjectWrapper[] {
```
atruskie, line 327: that's a weird name?

can we make currentSubSelection just behave like this method?

or maybe name this method `effectivelySelectedSubjects`

src/components/verification-grid/verification-grid.ts
```diff
@@ -1598,6 +1615,54 @@ export class VerificationGridComponent extends WithShoelace(AbstractComponent(Li
...
+    // We call "requestUpdate" on the tile so that the progress meter is updated
+    // with its new required decisions.
+    // I don't like manually calling "requestUpdate" here, but the component is
+    // an internal component that we control, so while this is an indication of
+    // bad state updates, it's not user facing, so I have just done a dirty
+    // requestUpdate here.
+    // TODO: Correctly fix the update triggers in the tile component so that we
+    // don't have to manually request an update.
+    tile.model = subject;
+    tile.requestUpdate();
```
atruskie, line 1663: two alternate patterns to consider:

- decisions required in a subject could be a signal. then the tile would know when to update if the signal is being monitored
- we could use a middle-man method on the tile (something like `updateSubject` or `updateSubjectDecision` which could trigger an internal update request

src/helpers/themes/decisionColors.ts
```diff
@@ -25,14 +43,30 @@ const classificationColors = `
     ).join("")}
 `;
 
+const tagPromptColors = `
+  --correction-true: ${hatchedBackground("#dd0")};
+  --correction-false: ${hatchedBackground("#dd0")};
+`;
+
```
atruskie, line 50: I don't think the tag correction should be hatched - it's not really that special. It's like a verification or classification.

src/helpers/themes/globalStyles.css
```diff
@@ -148,6 +148,15 @@ a {
...
 }
 
+del {
+  text-decoration: line-through;
+  text-decoration-color: var(--oe-danger-color);
+}
+
+ins {
+  text-decoration: none;
+}
```
atruskie, line 158: maybe trial a style here? light blue (info?) underline? 

theme colour for font?

src/helpers/themes/globalStyles.css
```diff
@@ -221,6 +230,27 @@ dialog {
   }
 }
 
+[popover]:not(dialog) {
+  position: relative;
+
+  /* Make the popover use the animation where it grows/shrinks when shown/hidden */
+  animation:
+    scale-in 0.1s forwards,
+    fade-in 0.2s forwards;
```
atruskie, line 239: not a fan of random magic animations: should  be variables so they're consistent when we start using them more?

not required for this pr

src/helpers/attributes.ts
```diff
@@ -43,16 +43,30 @@ export const tagConverter = (value: string | null): Tag => {
  * </oe-spectrogram>
  * ```
  */
-export const callbackConverter = (value: string | ((...params: any) => any)) => {
+export const callbackConverter = <T>(value: string | null | ((...params: any[]) => T)) => {
```
atruskie, line 46: not keen on the existence of this. I assume this is here so we can convert string attributes to functions... maybe we just don't allow that, even if it does look nice.

not requiring this to be changed

atruskie (changes requested):
Great work.

I think we should include a default MRU implementation. I'm happy to make it weak (i.e. the MRU only persists for the current page load) and overridable (maybe in a later PR - we could add an MRU as slotted tag items). But the important thing is something basic just works.

I don't think the tag correction task should use our special hatching color scheme. Should just be an ordinary color

src/components/verification-grid/verification-grid.ts
hudson-newey: I think something like `updateSubjet` would work best for now

src/helpers/attributes.ts
hudson-newey: Yeah, this is so that you can write javascript directly in html attributes

e.g.

```html
<oe-tag-prompt
	search="() => [{ text: 'test' }]"
></oe-tag-prompt>
```

atruskie: yeah this is a very weak dislike. leave it as is... just smells and I'm not sure why

atruskie:
Ok we talked about this offline a bit: I think we should remove all mention of _correction_.

`<oe-tag-prompt>` is meant to allow a user to supply a new preferred tag. This is a correction when there is an existing tag, but it also simply an application of a tag (not a correction) when there is no existing tag.


ok: for the local download this is my executive decision
 
- existing columns: tag = 'crow'
- oe_tag: 'crow'
- oe_verified: 'false'
- oe_new_tag: 'lorikeet'

> And if there is a missing oe_tag from the original data source, we leave the oe_tag cell blank in the download

> Changes that I got from this conversation: I'll rename tagCorrection to newTag and make no functionality changes to how tag correction works
 
yup, but also, i think a bunch of internal renames too

> Yeah, and some copy changes too

src/components/decision/tag-prompt/tag-prompt.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    this.close();
+  }
+
+  private popoverTemplate(): HTMLTemplateResult {
+    return html`
+      <dialog id="tag-popover" popover @toggle="${this.handleToggle}">
+        <div class="tag-popover-header">
+          <h3 class="tag-popover-title">Tag Correction</h3>
+          <button class="tag-popover-close oe-btn-secondary oe-btn-small" @click="${this.close}">x</button>
```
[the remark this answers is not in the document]
hudson-newey: https://github.com/ecoacoustics/web-components/pull/435/commits/8c456e5573daba3c8914b3e53ffba83e0be9b9a9

src/components/verification-grid-tile/verification-grid-tile.ts
```diff
@@ -59,13 +63,15 @@ const shortcutTranslation = {
...
  *
+ * @event selected
  * @event tile-loaded
  */
 @customElement("oe-verification-grid-tile")
 export class VerificationGridTileComponent extends SignalWatcher(WithShoelace(AbstractComponent(LitElement))) {
   public static styles = [unsafeCSS(verificationGridTileStyles), decisionColors];
 
   public static readonly selectedEventName = "selected";
+  public static readonly loadedEventName = "tile-loaded";
```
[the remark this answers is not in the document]
hudson-newey: https://github.com/ecoacoustics/web-components/pull/435/commits/a1f24eb94d7d50aabd35994ad1553b3c6a958b74

src/components/verification-grid/verification-grid.ts
hudson-newey: I think making `this.currentSubSelection` behave like this getter is more intuitive.

I like the idea

src/helpers/themes/globalStyles.css
hudson-newey: Mock up with info color underline:

image

I'm not sure if the underline provides much benefit without the tooltip, and it does have a drawback of adding more color & noise to the verification grid.

---

image

I think I personally like using `var(--oe-success-color)` as the font color.
And I think that this is consistent with other del + ins design I've seen across other websites e.g. GitHub

```diff
- Noisy Miner
+ Brush Turkey
```

src/components/decision/tag-prompt/tag-prompt.ts
```diff
@@ -0,0 +1,144 @@
...
+  public search: TypeaheadCallback<Tag> = () => [];
+
+  @query("#tag-popover")
+  private readonly tagPopover!: HTMLDialogElement;
+
+  @query("#tag-typeahead")
+  private readonly tagTypeahead!: TypeaheadComponent;
+
+  public get decisionModels(): Partial<DecisionModels<Decision>> {
+    throw new Error("Method not implemented.");
```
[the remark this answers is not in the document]
hudson-newey: This comes as a required method from the abstract `DecisionComponent`.
End users shouldn't be modifying this method.

I'll do a bit of digging, because I'm not sure if it's needed anymore.

I originally used it in the bootstrap to get the decision colors.
However, now that I'm exporting the constructor, I think I can use that.

Using the `Decision` constructor is also less opinionated because it doesn't require us to use mock data for stuff like the `Verification.confirmed` field

src/components/decision/tag-prompt/tag-prompt.ts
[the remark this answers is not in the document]
atruskie: yeah perfect - by svg i really meant shoelace icon (which is svg underneath)

src/helpers/themes/globalStyles.css
atruskie: yeah it might be too much. But I like your suggestion. 

I think it's useful to visibly distinguish a correct. Let's test it with users and see who complains

src/components/decision/tag-prompt/tag-prompt.ts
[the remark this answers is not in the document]
hudson-newey: recap: We only need this function for the classification bootstrap, so the colors are correct.

atruskie (approved):

PR #530, opened 2025-09-19
fix: Use "audacity" as default color map

hudson-newey:
# fix: Use "audacity" as default color map

## Final Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,6 @@ Full page example:
           class="large"
           src="/example.flac"
           window-size="1024"
-          color-map="audacity"
           scaling="stretch"
         ></oe-spectrogram>
       </oe-indicator>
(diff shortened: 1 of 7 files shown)


atruskie (approved):

atruskie:
A little bit of context for historical notes:

_I_ thought our default colour map was always `audacity`. I was wrong.

@hudson-newey always thought it was meant to be `grayscale`.

We first deployed these components with the `<verification-grid />` which also overrode the defaults to `audacity` - which further reinforced by erroneous assumption. There was some offline discussion for choosing colour maps which we didn't document so what happened there was unclear.

Our reasoning is as follows:

- colour spectrograms are more appealing, and appeal more to novice users
- expert users tend to express a preference for grayscale spectrogram due their uniform perceptibility of intensity. 
- the audacity colour map is uniformly perceptual so it is a good default for both users:
    -  since it has mass appeal
    - and it technically accurate (uniformly perceived) 
    - and expert users tend to be fewer than novice users (so better to appeal to larger crowd)
    - and expert users are more resilient to different presentations, but novices are not - so it is good to default to an appealing choice
- and lastly, we expect a good settings implementation will allow customisability anyway, even at a site-wide level

Issue #563, opened 2025-10-24
Conditionally render media control settings


hudson-newey:
I've done some testing and found that one of our major bottlenecks is the shoelace submenu which is extremely slow.

<img width="2252" height="1324" alt="Image" src="https://github.com/user-attachments/assets/ce305011-d76d-496b-bea7-bdfd5ef25716" />

We can hack around this by conditionally creating the submenu's when the settings menu is opened/closed.

PR #564, opened 2025-10-24
fix: Media controls & grid resizing performance

hudson-newey:
# fix: Media controls & grid resizing performance

## Changes

- Conditionally render the media controls setting menus to reduce shoelace calls to getComputedStyle
- Cache axes number & title sizes
- Use resizeObserver to update spectrogram container size instead of using getBoundingClientRect which causes reflow
- Disable alpha channel during spectrogram rendering for to improve compute time
- Disable image smoothing on spectrogram rendering for crisper spectrograms
- Correctly destroy source & destination surfaces on spectrogram worker destruction
- When computing grid size target of N, only use 1 flatten instead of flattening n times
- When computing factors, only search up to sqrt(target) by producing two results per factor (by flipping rows & columns for second result)
- Use bitwise & for even check in loop directive

## Visual Changes

**Videos have been exaggerated with 4x CPU throttling to mimic a mid-low end machine**

[Screencast_20251024_211841.webm](https://github.com/user-attachments/assets/d8396d6b-edbd-47f3-8ded-722d78ea0521)

_Before performance optimizations_

[Screencast_20251024_212538.webm](https://github.com/user-attachments/assets/a769dbf7-8421-41ee-8871-e9bc499bb59e)

_After performance optimizations_

## Related Issues

Fixes: #563

## Final Checklist

### DIFF
--- a/src/helpers/directives.ts
+++ b/src/helpers/directives.ts
@@ -44,9 +44,9 @@ export function* loop<T>(
     let i = 0;
 
     for (const value of items) {
-      const first = i == 0;
-      const last = i == items.length - 1;
-      const even = (i & 1) == 0;
+      const first = i === 0;
+      const last = i === items.length - 1;
+      const even = (i & 1) === 0;
       const odd = !even;
 
       yield f(value, { index: i++, first, last, even, odd });


src/components/axes/axes.ts
```diff
@@ -125,6 +125,13 @@ export class AxesComponent extends SignalWatcher(ChromeProvider(LitElement)) {
   // I have this as a private property so that we don't have to re-calculate the
   // value every time we need to use it
   private emUnitFontSize!: Size;
+
+  /** A font size determined by getting the font size of the "0.0" string */
+  private numberFontSize!: Size;
```
atruskie: good doc, but why is this concept added?

src/components/axes/axes.ts
```diff
@@ -245,8 +263,12 @@ export class AxesComponent extends SignalWatcher(ChromeProvider(LitElement)) {
...
     canvasSize: Readonly<Size>,
   ) {
-    const xTitleFontSize = this.calculateFontSize(this.xTitle);
-    const yTitleFontSize = this.calculateFontSize(this.yTitle);
+    this.cachedXTitleSize ??= this.calculateFontSize(this.xTitle);
+    this.cachedYTitleSize ??= this.calculateFontSize(this.yTitle);
+
+    const xTitleFontSize = this.cachedXTitleSize;
+    const yTitleFontSize = this.cachedYTitleSize;
+
```
atruskie: at the minimum the comment is now out of date,


but the comment's suggestion is good. Why don't we do that?

src/components/media-controls/media-controls.ts
atruskie: on the whole, i'm not supportive of this change.

This is being changed because of another issue - high frequency rendering of the components - which is not resolved AFAIK? Which makes this a red herring.

But if we were to think of another solution here: maybe we make the sub-menu once, and attach it to the menu when open?

Or maybe we patch shoelace's get computed style method to disable it?
Or file an issue with shoelace?
Or link to the faulty code just for documentation's sake?

src/helpers/controllers/dynamic-grid-sizes.ts
```diff
@@ -133,22 +135,36 @@ export class DynamicGridSizeController<Container extends HTMLElement> {
     const candidateShapes: GridShape[] = [];
 
     // find all the factors of the target grid size
-    for (let i = 1; i <= target; i++) {
+    const limit = Math.ceil(Math.sqrt(target));
```
atruskie: ok this is pretty smart

src/helpers/directives.ts
```diff
@@ -44,9 +44,9 @@ export function* loop<T>(
     let i = 0;
 
     for (const value of items) {
-      const first = i == 0;
-      const last = i == items.length - 1;
-      const even = i % 2 == 0;
+      const first = i === 0;
+      const last = i === items.length - 1;
+      const even = (i & 1) === 0;
```
atruskie: if you can prove this makes any difference we'll accept it

atruskie (changes requested):
Primarily: too many changes introduced into one PR.
Secondarily: not measurements of performance

PR #565, opened 2025-10-26
fix: Lazily transfer offscreenCanvas

hudson-newey:
# fix: Lazily transfer offscreenCanvas

Transfering the offscreen canvas causes a reflow.
By using RAF, we can transfer the canvas at an optimal time to maintain
60fps.

Note that we still start processing the audio + FFT, but transfer the
canas lazily.

If the worker completes the FFT compute in < 1 frame, the canvas will
immediately be painted to once transfered.
I set a "paint" flag instead of waiting so that we don't have to spin
and save compute resources.

## Changes

- Lazily transfer offscreen canvas to reduce non-optimal reflow

## Related Issues

NA

## Final Checklist

### DIFF
--- a/src/helpers/audio/worker.ts
+++ b/src/helpers/audio/worker.ts
@@ -166,6 +166,14 @@ function setupDestinationCanvas(canvas: OffscreenCanvas): void {
     alpha: false,
   }) as OffscreenCanvasRenderingContext2D;
   destinationSurface.imageSmoothingEnabled = false;
+
+  // Because the worker can finish generating the spectrogram before the main
+  // thread can transfer the canvas, we may have a buffered image to draw before
+  // the canvas was available.
+  // In this case, we immediately draw the spectrogram to the canvas.
+  if (hasBufferedImage) {
+    drawSpectrogramOntoDestinationCanvas(state.generation);
+  }
 }
 
 function regenerate(data: GenerationMetadata): void {


src/helpers/audio/worker.ts
```diff
@@ -26,6 +26,7 @@ let spectrogram: SpectrogramGenerator;
 let options: SpectrogramOptions;
 
 let state: WorkerState;
+let hasBufferedImage = false;
```
atruskie: i'm in favour of removing this - also doc missing on declaration

src/helpers/audio/worker.ts
```diff
@@ -121,25 +134,48 @@ function drawSpectrogramOntoDestinationCanvas(_generation: number): void {
...
   state.workerReady();
 }
 
+function setupDestinationCanvas(canvas: OffscreenCanvas): void {
+  destinationCanvas = canvas;
+
+  destinationSurface = destinationCanvas.getContext("2d", {
+    alpha: false,
+  }) as OffscreenCanvasRenderingContext2D;
+  destinationSurface.imageSmoothingEnabled = false;
```
atruskie: this is a secondary change in an optimization PR, it does not belong here.

You're right, each pixel is meaningful, but in our source buffer, not in our destination canvas. Because we have stretch and natural sizing we want to take advantage of image smoothing. Only on the original size, or in sizes that are exact scaled intervals of the original size do we probably want to disable this.

We did actually test this when we first set up the project.

We can revisit, in another PR, with before and after shots, for various sizes and sizing modes

src/helpers/audio/worker.ts
```diff
@@ -121,25 +134,48 @@ function drawSpectrogramOntoDestinationCanvas(_generation: number): void {
...
+
+  // Because the worker can finish generating the spectrogram before the main
+  // thread can transfer the canvas, we may have a buffered image to draw before
+  // the canvas was available.
+  // In this case, we immediately draw the spectrogram to the canvas.
+  if (hasBufferedImage) {
+    drawSpectrogramOntoDestinationCanvas(state.generation);
+  }
+}
```
atruskie: why not just always call drawSpectrogramOntoDestinationCanvas and remove the need for a global state?

hudson-newey: There might be nothing to draw if the FFT hasn't been computed yet

src/helpers/audio/worker.ts
```diff
@@ -104,6 +105,18 @@ function renderImageBuffer(buffer: Uint8ClampedArray, generation: number): void
...
+  // If we attempt to draw to the canvas before it has been transferred, we set
+  // a hasBufferedImage flag which indicates that when we receive the
+  // "transfer-canvas" message, we should immediately draw to the canvas.
+  //
+  // Because we only transfer the canvas on frame updates, this means that if
+  // worker the worker finishes generating the spectrogram before the next
+  // frame, we will not have a canvas to draw to yet.
+  if (!destinationCanvas || !destinationSurface) {
+    hasBufferedImage = true;
+    return;
```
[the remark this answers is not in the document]
atruskie: >But that means that the worker is opinionated about how it's consumed used which breaks encapsulation.

I don't know what you mean

> I've seen a couple comments now. Why are you against the flag?

Because it smells like a leaky abstraction.


--- 

My solution is this: if you don't have a destination surface to paint to, schedule a paint on the next opportunity (thinking about this, rAF would work here too). There's no global state.

simply:


```
  // frame, we will not have a canvas to draw to yet.
  if (!destinationCanvas || !destinationSurface) {
    // so try again next frame
    requestAnimationFrame(() => drawSpectrogramOntoDestinationCanvas(_generation);
    return;
}
```

I'm not sure how we ended up with "but making the workers internal implementation dependent on how its called seems even worse."

src/helpers/audio/worker.ts
atruskie: partly i don't like more global state - especially with the message based flow that happens with this code

Also it seems more elegant to me to do a callback (https://github.com/ecoacoustics/web-components/pull/565#discussion_r2464418625). But that's subjective.

